PR #31, opened 2014-09-28
PlatformIO Library Registry manifest file

ivankravets:
* This library in Web: http://platformio.ikravets.com/#!/lib/show/83/RTC
* Specification: [PlatformIO Library Manager](http://docs.platformio.ikravets.com/en/latest/librarymanager/index.html)
* Integration: [IDE Integration](http://docs.platformio.ikravets.com/en/latest/ide.html)

### DIFF
--- a/library.json
+++ b/library.json
@@ -1,10 +1,12 @@
 {
-  "name": "Arduino-RTC",
+  "name": "RTC",
   "keywords": "time, rtc",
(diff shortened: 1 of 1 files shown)


ladyada:
hiya, please use the arduino platform file. thats all we can do for now. thx!

PR #34, opened 2014-12-13
Fix overflow in timespan constructor. Cast to int32_t is happening too late

murmlgrmpf:
The timespan object is exhibiting an overflow if hours>8 are used. This should not be the case and can be fixed by casting (defining) all input parameters to int32_t. This does not increase the memory usage, as the final result _seconds needs to be casted to int32_t anyway.


ladyada:
apologies for the delay, i believe
https://github.com/adafruit/RTClib/commit/03d0a6ad291dd2968aaf0127e122687b5ece7829
resolves this!

PR #39, opened 2015-05-06
Support 86Duino board

acen2009:
Hello, now the Adafruit RTC library also supports the 86Duino board (another Arudino compatible board) on this branch :)


ladyada:
without merging, i believe that https://github.com/adafruit/RTClib/commit/852a10239706f3217fc8e83c74e3b7b8049fa874 does it

PR #41, opened 2015-06-13
Add support for DS3231 in backwards-compatible way

davidb24v:
By moving a few defines to the header file and adding an optional argument with default value to the square wave routines, this library can also support the DS3231. To use with DS3231, just pass 0x0E to the square wave routines and it will use that as the control register address. All existing DS1307 code is unaffected.

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -12,13 +12,6 @@
  #define WIRE Wire1
 #endif
 
-#define DS1307_ADDRESS  0x68
-#define DS1307_CONTROL  0x07
-#define DS1307_NVRAM    0x08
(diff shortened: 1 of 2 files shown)


ladyada:
pardon the delay in replying. perhaps it would make more sense to make a new subclass so that it has a new name? most of the code is identical, could be a fairly simple copy & paste + mod...

davidb24v:
Sure, I'll rework this and push the changes when I build another clock (you can never have enough clocks). I want to do some more testing anyway as it failed a couple of times on me recently. It won't be quick but I should get onto it this year some time...

PR #55, opened 2015-11-19
Updated RTCMillis class

arkhipenko:
- exact same interface between RTC_DS1307 and RTCMillis classes, so only initial declaration would vary, the rest of the code could remain the same:
  # ifndef _USERTCMILLIS_
  
  RTC_DS1307 rtc; 
  # else
  
  RTC_Millis rtc;
  # endif
- millis() rollover fix so RTCMillis could be used for over 43 days.

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -324,14 +324,37 @@ void RTC_DS1307::writenvram(uint8_t address, uint8_t data) {
 ////////////////////////////////////////////////////////////////////////////////
 // RTC_Millis implementation
 
-long RTC_Millis::offset = 0;
+boolean RTC_Millis::begin(void) {
+  offset = 0;
+  adjust(DateTime(__DATE__,__TIME__));
+  return true;
+}
 
 void RTC_Millis::adjust(const DateTime& dt) {
     offset = dt.unixtime() - millis() / 1000;
+	prevMillis = millis();
+    countRollovers = 0;
 }
 
 DateTime RTC_Millis::now() {
-  return (uint32_t)(offset + millis() / 1000);
+  checkRollover();
+  return (uint32_t)(offset + millis() / 1000 + (countRollovers * 4294967L) + (countRollovers*296/1000) );
 }
 
+// checkRollover should be run periodically to ensure a rollover is captured.
+// Since rollovers happen once in ~43 days, "periodically" could be once a day, once a week, or even once a month!
+void RTC_Millis::checkRollover() {
+	if (prevMillis > millis()) countRollovers++;
+	prevMillis = millis();
+}
+
+
+Ds1307SqwPinMode RTC_Millis::readSqwPinMode() {
+  int mode;
+
+  mode = 0x80 | 0x12;
+
+  mode &= 0x93;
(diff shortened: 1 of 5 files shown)


edgar-bonet:
I see a few minor issues in this pull request:

1. `RTC_Millis::now()` calls `millis()` three times (once directly and twice through `checkRollover()`)
2. when `millis()` first rolls over, `RTC_Millis` gets late by 296&nbsp;ms
3. `RTC_Millis` uses more CPU and RAM than needed
4. the pull request is too wide in scope

The first point has no practical consequence. Yet, since the different calls are not guaranteed to give the same result, as a matter of consistency `now()` should call `millis()` only once.

The second point means that the two second “tics” on each side of the rollover are separated by 1.296&nbsp;seconds instead of one second. This is because `countRollovers*296/1000` involves an integer division that yields zero. Note, however, that on the _fourth_ rollover the clock gets _fast_ by 704&nbsp;ms, so the error is not cumulative. A naive fix would be to replace

```c++
millis() / 1000 + countRollovers*296 / 1000
```

with

```c++
(millis() + countRollovers*296) / 1000
```

but this incurs the risk of the addition overflowing.

As for the fourth point, by “wide scope” I mean that fixing the rollover problem and enriching the API are unrelated issues that deserve separate pull requests.

I just submitted another pull request (#92) addressing all the points above.

arkhipenko:
This is indeed more optimized code. Thank you for taking the time and making it more efficient. I never had the time to go back and do it properly. 

However,
I would recommend keeping begin() method consistent across all classes so the software implementation can be swapped out without subsequent changes to the code.
Also, is there a reason why lastMillis and lastUnix are static? I get it, it is unlikely that anyone would use more than one RTC_Millis object, but if they do for whatever reason, this will produce interesting results...

edgar-bonet:
> I would recommend keeping begin() method consistent across all classes [...]

This looks like a very reasonable recommendation to me.

> is there a reason why lastMillis and lastUnix are static?

Two reasons actually:

1. to be consistent with the fact that `offset` is static in the current base branch
2. it is required because all the methods of the class are static

Of course, you can make everything in the class non-static, and you even have a perfectly good rationale for doing so, but this does not help with the millis rollover. Allowing the user to instantiate multiple `RTC_Millis` objects is certainly a valid concern, but one that is completely _unrelated_ to the rollover issue.

This all boils down to my “too wide scope” point above. Most code reviewers like narrowly focused pull requests: one issue → one pull request → one discussion. It helps code review and it helps in keeping the discussion focused. I have no experience with the Adafruit folks in particular, but judging from their PR backlog, I would assume that anything that makes the code review easier for them is welcome.

For reference, see these two pull request style guides:

* [The Art of Pull Requests][the-art], by Danny Preussler, has “keep PRs small” as its very first recommendation
* [10 tips for better Pull Requests][10-tips], by Mark Seemann, has “Make it small” and “Do only one thing” as its first two tips.

[the-art]: https://hackernoon.com/the-art-of-pull-requests-6f0f099850f9
[10-tips]: http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests/

drak7:

PR #65, opened 2016-05-11
Added support for the PCF8563 and an example

goexp:
I added support for the PCF8563 based on the PCF8523 implementation that was present in the library


goexp:
Added alarm support for the PCF8563 as well

drak7:

PR #78, opened 2017-02-21
DS1339 support

odaki:
add code for DS1339

-  I have add a new class and a enum to support DS1339 RTC that based on DS3231 code.

- To treat DS1339's alarm function is not yet implemented.

-  I  wrote examples to helps with test process.

Thank you.


drak7:

PR #84, opened 2017-11-08
Update ds3231.ino

lauriemustafic:
I've just adjusted the comment to correctly and more clearly explain what TimeSpan() accepts and what the example values are doing.

There are no actual code-based changes but I think for a novice, the suggest comment change is less confusing than what's currently there. 

Currently, the comment says it will add 7 days and 30 seconds to the time, but it actually add 7 days, 12 hours, 30 minutes and 6 seconds.

I probably actually should've adjusted [line 62 Serial.print()](https://github.com/adafruit/RTClib/pull/84/files#diff-8a70ea90e6a614d965654ed3ad68b2afR62) to reflect the alteration.

### DIFF
--- a/examples/ds3231/ds3231.ino
+++ b/examples/ds3231/ds3231.ino
@@ -55,7 +55,8 @@ void loop () {
     Serial.print(now.unixtime() / 86400L);
     Serial.println("d");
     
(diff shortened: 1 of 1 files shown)


drak7:
Good catch, adjust line 62 as well and we can merge this.

PR #89, opened 2018-01-09
added getAlarm and setAlarm

carlfarrington:
Hi. I just added a simple thing to be able to set/get the alarm from the PCF8523.
I have the Feather M0 and the adalogger shield. I wanted to store some configuration flags and there's no eeprom on this board so the alarm seems like a reasonable thing to use.

I am only just learning C / C++ so apologies if the code is terrible.


drak7:

PR #92, opened 2018-06-05
Make RTC_Millis immune to millis() rollover events

edgar-bonet:
Every 49.7&nbsp;days (2<sup>32</sup>&nbsp;ms), `millis()` rolls over to zero. When this happens, the `RTC_Millis` “soft RTC” jumps back in time by 49.7&nbsp;days.

This pull request fixes the issue, provided that `RTC_Millis::now()` is called more often than the rollover period. The code relies on the fact that, owing to the rules of [modular arithmetic][], computing `millis() - lastMillis` [is rollover safe][safe].

Here is a test sketch coercing `millis()` to roll over early:

```c++
#include <Wire.h>
#include <RTClib.h>

void setup() {
    extern unsigned long timer0_millis;
    timer0_millis = -1500;  // 1.5 seconds before rollover
    RTC_Millis::begin(DateTime(__DATE__, __TIME__));
    Serial.begin(9600);
    Serial.println(" millis()    RTC_Millis::now()");
    Serial.println("-------------------------------");
}

void loop() {
    char buffer[48];
    DateTime t = RTC_Millis::now();
    sprintf(buffer, "%10lu  %04d-%02d-%02d %02d:%02d:%02d", millis(),
        t.year(), t.month(), t.day(), t.hour(), t.minute(), t.second());
    Serial.println(buffer);
    delay(1000);
}
```

The output produced by an Arduino Uno, without this fix:

```text
 millis()    RTC_Millis::now()
-------------------------------
4294965796  2018-06-05 20:51:09
4294966830  2018-06-05 20:51:10
       535  2018-04-17 03:48:24
      1536  2018-04-17 03:48:25
```

And the output with the fix applied:

```text
 millis()    RTC_Millis::now()
-------------------------------
4294965796  2018-06-05 20:51:27
4294966830  2018-06-05 20:51:28
       534  2018-06-05 20:51:29
      1536  2018-06-05 20:51:30
```

As a side effect of the changed algorithm, the “soft RTC” can now be set with sub-second resolution: once the time is set with `adjust()`, it will not change until a full second has elapsed. In a sense, the “implicit milliseconds field” of the time is cleared by `adjust()`, whereas in the current master branch it is always `millis()%1000`.

I am aware that this rollover issue has been addressed by pull request #55. However, the pull request I am submitting here is more efficient (fewer changes in the code, lower CPU and RAM usage) and more focused on this particular issue.

[modular arithmetic]: https://en.wikipedia.org/wiki/Modular_arithmetic
[safe]: https://arduino.stackexchange.com/a/12588/7508

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -337,14 +337,26 @@ void RTC_DS1307::writenvram(uint8_t address, uint8_t data) {
 ////////////////////////////////////////////////////////////////////////////////
 // RTC_Millis implementation
 
-long RTC_Millis::offset = 0;
+// Alignment between the milis() timescale and the Unix timescale. These
+// two variables are updated on each call to now(), which prevents
+// rollover issues. Note that lastMillis is **not** the millis() value
+// of the last call to now(): it's the millis() value corresponding to
+// the last **full second** of Unix time.
+uint32_t RTC_Millis::lastMillis;
+uint32_t RTC_Millis::lastUnix;
 
 void RTC_Millis::adjust(const DateTime& dt) {
-    offset = dt.unixtime() - millis() / 1000;
+  lastMillis = millis();
+  lastUnix = dt.unixtime();
 }
 
+// Note that computing (millis() - lastMillis) is rollover-safe as long
+// as this method is called at least once every 49.7 days.
 DateTime RTC_Millis::now() {
-  return (uint32_t)(offset + millis() / 1000);
+  uint32_t elapsedSeconds = (millis() - lastMillis) / 1000;
+  lastMillis += elapsedSeconds * 1000;
+  lastUnix += elapsedSeconds;
+  return lastUnix;
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/RTClib.h
+++ b/RTClib.h
@@ -121,7 +121,7 @@ public:
 };
 
 // RTC using the internal millis() clock, has to be initialized before use
-// NOTE: this clock won't be correct once the millis() timer rolls over (>49d?)
+// NOTE: this is immune to millis() rollover events
 class RTC_Millis {
 public:
     static void begin(const DateTime& dt) { adjust(dt); }
@@ -129,7 +129,8 @@ public:
     static DateTime now();
 
 protected:
-    static long offset;
+    static uint32_t lastUnix;
+    static uint32_t lastMillis;
 };
 
 #endif // _RTCLIB_H_


shantanoo-desai:
@edgar-bonet Quick question.

If i use your PR on a board and use `t.unixtime()` for `Datetime t` will the precision also be in milliseconds?

edgar-bonet:
@shantanoo-desai: This PR fixes a rollover issue. It has no effect on the precision:

* `DateTime::unixtime()` has a resolution of one second. It will not give you millisecond-level time stamps.
* The elapsed time between successive “ticks” of `unixtime()` should be about 1,000±1&nbsp;ms. The 1&nbsp;ms jitter comes from the Arduino implementation of `millis()`, which is _not_ accurate to the millisecond.
* The long-term accuracy will likely be poor, especially if your Arduino is clocked off a ceramic resonator. Expect a drift of the order of ±0.1% (more than one minute per day).
* The drift rate is not even going to be consistent. Expect random variations and high sensitivity to the ambient temperature.

See [Arduino clock frequency accuracy][jorisvr] for more information about the kind of accuracy you can expect from the Arduino's internal clock. In summary: if you need an accurate clock, use a real RTC. Alternatively, use an Arduino clocked off a real quartz crystal and correct the (almost constant) drift in soft.

I have implemented a variation of `RTC_Millis` that can be finely trimmed in order to remove the drift. I call it `RTC_Micros`. It can be significantly better than `RTC_Millis` _if_ you take the time to measure and correct the drift. It is obviously still susceptible to the random frequency variations of the Arduino's internal clock, so it should perform way better on a crystal-clocked Arduino. Tell me if you are interested, in which case I will submit this as another pull request.

[jorisvr]: http://jorisvr.nl/article/arduino-frequency

shantanoo-desai:
@edgar-bonet is there anyway I can club the millisecond timestamps to the unix epoch time. I wish to obtain some sensor values which are annotated with timestamps (epoch timestamps) but I want to the precision to be in milliseconds.

shantanoo-desai:
@edgar-bonet really quick question: how do I get the difference in ms acc. to your patch?

```
RTC_DS3231 rtc;
RTC_Millis software_time;

void setup() {
  // put your setup code here, to run once:
  Wire.begin(I2C_SDA, I2C_SCL, 400000);
  Serial.begin(9600);
  if (!rtc.begin()) {
    Serial.println("No RTC Detected");
  }
  // adjust RTC Time
  // adjust software timer
  software_time.adjust(rtc.now());

}

void loop() {
  // put your main code here, to run repeatedly:
  DateTime rtc_now_time = rtc.now();
  DateTime software_now_time = software_time.now(); // what goes here?
  Serial.println();
  Serial.print("RTC now: "); Serial.print(rtc_now_time.unixtime());
  Serial.println();

  delay(1000);

}
```

edgar-bonet:
@shantanoo-desai: This is not the right venue for discussing issues that, at this point, are getting _quite far_ from the original purpose of the pull request. You may try <https://arduino.stackexchange.com/>.

edgar-bonet:
Just rebased on top of the current master. Is anyone still maintaining this repo?

ladyada:
we're focusing on some other Repos that have bugs in them, we'll get to this one at some point!

PR #101, opened 2018-10-02
Update RTClib.h

PeterEmbedded:
Code formatting consistency.
No actual code is changed, so nothing can possibly break.


ladyada:
ok thanx!

Issue #99, opened 2018-08-28
Warning when compiling


MarioJose:
Code working, but get warning when compiling

/home/user/.platformio/lib/RTClib_ID83/RTClib.cpp: In constructor 'DateTime::DateTime(uint32_t)':
/home/user/.platformio/lib/RTClib_ID83/RTClib.cpp:86:18: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
if (days < 365 + leap)
^

Phoenix1747:
Yeah that's just because you compare the ``uint16_t days` to a normal signed integer `365`. This shouldn't be a real problem at all. By the way, I don't get this warning, how do you use this function?

MarioJose:
My code:

```{c}
#include <Arduino.h>

#include <string.h>
#include <inttypes.h>

#include <SD.h>
#include <RTClib.h>

RTC_DS1307 rtc;
DateTime calendar;


void setup()
{
  Serial.begin(9600);

  calendar = rtc.now();

  Serial.println(calendar.minute());


}

void loop()
{

}

```

And Atom+PlatformIO build output:

```
Processing uno (platform: atmelavr; board: uno; framework: arduino)
--------------------------------------------------------------------------------
Verbose mode can be enabled via `-v, --verbose` option
PLATFORM: Atmel AVR > Arduino Uno
SYSTEM: ATMEGA328P 16MHz 2KB RAM (31.50KB Flash)
Library Dependency Finder -> http://bit.ly/configure-pio-ldf

LDF MODES: FINDER(chain) COMPATIBILITY(soft)
Collected 36 compatible libraries
Scanning dependencies...
Dependency Graph
|-- <RTClib> 1.2.1
|   |-- <Wire> 1.0
|-- <SD> 1.2.2
|   |-- <SPI> 1.0
Compiling .pioenvs/uno/src/main.cpp.o
Compiling .pioenvs/uno/lib741/Wire/Wire.cpp.o
Compiling .pioenvs/uno/lib741/Wire/utility/twi.c.o
Compiling .pioenvs/uno/lib80d/RTClib_ID83/RTClib.cpp.o
Compiling .pioenvs/uno/lib234/SPI/SPI.cpp.o
/home/user/.platformio/lib/RTClib_ID83/RTClib.cpp: In constructor 'DateTime::DateTime(uint32_t)':
[...]

PR #109, opened 2019-02-18
Fix signed/unsigned compare warning bug

paynterf:
Fix for Issue 99 - warning about signed/unsigned comparison mismatch

This pull request is for the fix to Issue #99 Warning when compiling.  The warning is caused by 
comparing an unsigned int 'day' to a signed int ('365' which is regarded as an 'int' by default).
The fix is to replace '365' by '365U' in two places, making 365 an unsigned int.


drak7:
If you'd still like to merge this please rebase on top of the latest code and resubmit. Thanks!

PR #117, opened 2019-07-28
Add function to create custom date and time formats

NightSkySK:
Added function "format" to DateTime class. 
This function allows the user to easily define how the return string should look like. 

Configuratuion defined as:
    strncpy(buf,"DD.MM.YYYY  hh:mm:ss\0",100);
    Serial.println(now.format(buf));
Result --> 28.07.2019  15:09:10

    strncpy(buf,"YYMMDD\0",100);
    Serial.println(now.format(buf));
Result --> 190728

    strncpy(buf,"YY.MM.DD\0",100);
    Serial.println(now.format(buf));
Result --> 19.07.28


NightSkySK:
Can you please advice me which documentation I need create to clear error:
/home/travis/build/adafruit/RTClib/RTClib.h:60: warning: parameters of member DateTime::format are not (all) documented

Where I need to create function documentation/description?

edgar-bonet:
Interesting proposal. I have a few comments and questions:

1. Why the `\0` in the format strings?

   Literal strings in C++ are implicitly terminated by an ASCII NUL. A string such as `"DD.MM.YYYY  hh:mm:ss\0"` ends thus with _two_ consecutive NULs (one explicit, one implicit). Is there a reason for this? If so, it should be documented, as it is non-intuitive.

2. Why the call to `strncpy()` in the examples?

   It would seem more natural to me to write

   ```c++
   char buf[] = "DD.MM.YYYY  hh:mm:ss";
   Serial.println(now.format(buf));
   ```

3. The format string is overwritten

   It kind of makes sense to use the format string as a buffer, as it avoids dynamically allocating memory. This behavior is however very unintuitive and should at least be documented. Otherwise it is very tempting (though incorrect) to write

   ```c++
   Serial.println(now.format("DD.MM.YYYY  hh:mm:ss"));
   ```

4. Alternative approach

   Maybe something like this could be easier to use:

   ```c++
   now.setFormat("DD.MM.YYYY  hh:mm:ss");
   Serial.println(now);
   ```

   That could be achieved by having `DateTime` inherit from `Printable`, and implementing the virtual `printTo(Print&)` method. Then there is no need to allocate a writable buffer, which can lead to RAM savings.

   Or maybe even make the format a static property. This way it doesn't have to be redefined every time it is needed:

   ```c++
   // In setup:
   DateTime.setFormat("DD.MM.YYYY  hh:mm:ss");

   // In loop:
   Serial.println(rtc.now());
   ```

ladyada:
@NightSkySK https://learn.adafruit.com/the-well-automated-arduino-library/doxygen

NightSkySK:
> 1. Why the \0 in the format strings?

Thank you, you are right it seems to be not necessary, I will try to delete it and test it. Once it will not cause any unexpected behavior will be updated in the next pull request.

>2.  Why the call to strncpy() in the examples?

Indeed this part makes unnecessary complication of code. Sorry, I'm still a beginner. Let me update the code.

>4. Maybe something like this could be easier to use:

```
now.setFormat("DD.MM.YYYY  hh:mm:ss");
Serial.println(now);
```

I'm going to use different formats within one sketch, that's why I prefer to keep possible to call now.format() function in reference to two or more formats.  We can imagine that some authors would like to allow users to configure date format in some UI, or like im my case I need time stamps like [YYYY-MM-DD hh:mm:ss], but file name on SD card to be named as YYMMDD.log

edgar-bonet:
NightSkySK wrote:

> [`strncpy()`] is a necessary step [...] I've tried to set global variables

The buffer should not be a global array, because it would then be initialized only once, at the start of the program. It should be a local variable:

```c++
void loop() {
    DateTime now = rtc.now();
    char format1[] = "DD.MM.YYYY  hh:mm:ss";
    Serial.println(now.format(format1));
    char format2[] = "YYMMDD";
    Serial.println(now.format(format2));
    char format3[] = "YY.MM.DD";
    Serial.println(now.format(format3));
    delay(1000);
}
```

> I'm going to use different formats within one sketch

You still can use different formats with the “Printable” approach:

```c++
Serial.print("Date: ");
DateTime::setFormat("DD.MM.YYYY");
Serial.println(now);

Serial.print("Time: ");
DateTime::setFormat("hh:mm:ss");
Serial.println(now);
```

However, if you need the formatted string for something other than printing (e.g. a file name), then your approach makes more sense. I would however change the method name, the parameter name, and the documentation, in order to make crystal clear that the method is going to overwrite the format string:

```c++
char* DateTime::toString(char* buffer)
```

* Parameters:
  * `buffer`: array of `char` for holding the format description and the formatted `DateTime`. Before calling this method, the buffer should be initialized by the user with a format string, e.g. `"YYYY-MM-DD hh:mm:ss"`. The method will overwrite the buffer with the formatted date and/or time.
* Return value: a pointer to the provided buffer. This is returned for convenience, in order to enable idioms such as `Serial.println(now.toString(buffer));`.

Note that the method could be extended to formats such as “Sun Jul 28, 10:21 PM”. You may want to take a look at [Qt’s `QDateTime::toString()`][qt] for inspiration.

[qt]: https://doc.qt.io/qt-5/qdatetime.html#toString

NightSkySK:
@edgar-bonet Thank you for your kindly support. Let me check if I'm able to create a new pull request today evening, which will include the above suggestions.

NightSkySK:
New pull request contains renamed function toString, I've included more detailed function description in docs of the library as in example file. 

> Note that the method could be extended to formats such as “Sun Jul 28, 10:21 PM”. You may want to take a look at Qt’s QDateTime::toString() for inspiration.

This would need to completely rebuild the function behavior in most cases. I wish to take this challenge later if possible as the next stage.

edgar-bonet:
Two remarks about your implementation of the “DDD” and “MMM” formats:

1. You wrote

   ```c++
   if(buffer[i] == 'D' && buffer[i+2] =='D'){
   ```

   Before reading `buffer[i+2]` you should make sure `buffer[i+1]` is not NUL. Or, better yet, make sure it's a `'D'`. Otherwise you may end up reading (and worse, writing) past the end of the buffer. The same applies to the tests for the formats “MMM” and “YYYY”.

2. The `switch`/`case` constructs implementing the “DDD” and “MMM” formats can be avoided by storing the month and day names in arrays. For example:

   ```c++
   if (buffer[i] == 'D' && buffer[i+1] == 'D' && buffer[i+2] == 'D') {
     static PROGMEM const char day_names[] = "SunMonTueWedThuFriSat";
     const char *p = &day_names[3*dayOfTheWeek()];
     buffer[i] = pgm_read_byte(p);
     buffer[i+1] = pgm_read_byte(p+1);
     buffer[i+2] = pgm_read_byte(p+2);
   }
   ```

   The `PROGMEM` and `pgm_read_byte()` macros are an optimization that avoids keeping a copy of the array in RAM. It is specific to the AVR platform, but it seems non-AVR Arduino cores provide [a compatibility layer][pgmspace] that makes them safe to use in a cross-platform library.

   My test shows this approach saves 230&nbsp;bytes of flash compared to the `switch`/`case` constructs.

[pgmspace]: https://github.com/arduino/ArduinoCore-samd/blob/1.8.3/cores/arduino/avr/pgmspace.h

NightSkySK:
@drak7 @ladyada, Could you please review if that pull meet the Adafruit standards?

drak7:
Looks good! I'll merge it at this point, if you'd like to keep improving it feel free to submit a new PR.

NightSkySK:
@edgar-bonet Thank you for given support. I would not be able to complete it without your suggestions and hints. This was my first pull request to such library and I've learned a lot in this task.

Issue #114, opened 2019-06-18
[REGRESSION] DateTime corruption?


lovethat777:
DateTime causes problems.
The following code is self-explanatory: when _ct_ is declared, the _day_ and _month_ return invalid values:

EDIT: Tested a previous version (also labelled 1.2.1) which does not have this problem. so this is a regression.

```
#include "RTClib.h"      // https://github.com/adafruit/RTClib
#include <Time.h>        // https://github.com/PaulStoffregen/Time
#include <Timezone.h>    // https://github.com/JChristensen/Timezone


// If you declare "DateTime ct;", both month (13) and day (140) are invalid:
// 1560889639 2019-13-140 20:27:19

// If you UNCOMMENT "DateTime ct;" you get the following, which is correct:
// 1560889639 2019-6-18 20:27:19
//DateTime ct;

void setup () {
  Serial.begin(115200);

  time_t nowTime = 1560889639;

  Serial.print(nowTime);
  Serial.print(" ");
  Serial.print(year(nowTime));
  Serial.print("-");
  Serial.print(month(nowTime));
  Serial.print("-");
  Serial.print(day(nowTime));
  Serial.print(" ");
  Serial.print(hour(nowTime));
  Serial.print(":");
  Serial.print(minute(nowTime));
  Serial.print(":");
  Serial.println(second(nowTime));
}

void loop(){}
```

edgar-bonet:
I could not reproduce the problem on Linux with Arduino 1.8.5, using the current master branch of each of the libraries involved:

* RTClib: 5ddcf7c4d8aefa4286874426b158d7f3e0ec2b26 (Apr 13, 2019)
* Time: [d4fafb2][Time] (Jun 11, 2019)
* Timezone: [a913400][Timezone] (Nov 14, 2018)

Whether I include or not the line

```c++
DateTime ct;
```

I get the same result:

```text
1560889639 2019-6-18 20:27:19
```

[Time]: https://github.com/PaulStoffregen/Time/commit/d4fafb2c06c41a5ddf18f351bf30c5e22e1e7d2e
[Timezone]: https://github.com/JChristensen/Timezone/commit/a9134007b3c5a5274450474343bcabe06bc16292

lovethat777:
I used Arduino IDE 1.8.9 as well as 1.8.3 both on Windows (7 and XP) and deployed on a Pro Mini and a Nano, same results as mentioned. Also used the latest versions of RTClib, Time and Timezone (as you mentioned)

[...]

PR #125, opened 2019-10-09
Remove the last cell in the daysInMonth array

edgar-bonet:
This is a proposed fix for issue #114: an incompatibility with Paul Stoffregen's Time library arising from both libraries defining identical arrays of constants.

It has been suggested to fix the issue by _adding_ a dummy entry at the end of the array, in order to make it different from the one in the Time library. However, the issue can also be fixed by _removing_ the last item from the array. This array is used in computations where the total number of days from the previous months in the year is added or subtracted from a day count. Since December is never among the previous months, the twelfth cell in the array is not useful.

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -111,8 +111,12 @@ static void write_i2c_register(uint8_t addr, uint8_t reg, uint8_t val) {
 // utility code, some of this could be exposed in the DateTime API if needed
 /**************************************************************************/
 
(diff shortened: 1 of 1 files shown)


nobodyinperson:
I can confirm that it works when applied to my fork: https://github.com/nobodyinperson/RTClib/commit/fbd81fb42310036af249ae049743fd3306e977da

PR #130, opened 2019-12-22
ifdef pin definitions for ESP8266

### LINKED ISSUE #129: Documenting how to use with ESP8266?
If you use this with ESP8266 you need 

```c++
#include <Wire.h>
...
void setup () {
  Wire.begin(2, 14); // or some other pins
...
```

I was wondering if these two lines shouldn't be documented somewhere and/or be added as commented code to the example `.ino`s?

marcelstoer:
I did this for DS3231 only because it's the only one I have available for testing. Yet, it could/should be added to the other examples as well.

Fixes #129

### DIFF
--- a/examples/ds3231/ds3231.ino
+++ b/examples/ds3231/ds3231.ino
@@ -1,11 +1,17 @@
 // Date and time functions using a DS3231 RTC connected via I2C and Wire lib
 #include "RTClib.h"
+#ifdef ESP8266
+#include <Wire.h>
+#endif
 
 RTC_DS3231 rtc;
 
 char daysOfTheWeek[7][12] = {"Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"};
 
 void setup () {
+#ifdef ESP8266
+  Wire.begin(2, 14); // define your SDA & SCL pins
+#endif  
 
 #ifndef ESP8266
   while (!Serial); // for Leonardo/Micro/Zero


ladyada:
hiya @siddacious @marcelstoer thanks for the PR - we can't accept it because it will break usage for all Feather ESP users, we define the Wire pins in the board type definition. this stuff should be documented some other way, not changing the example :)

examples/ds3231/ds3231.ino
```diff
@@ -1,11 +1,17 @@
...
+#include <Wire.h>
+#endif
 
 RTC_DS3231 rtc;
 
 char daysOfTheWeek[7][12] = {"Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"};
 
 void setup () {
+#ifdef ESP8266
+  Wire.begin(2, 14); // define your SDA & SCL pins
```
marcelstoer: @ladyada could we address your comment by using the Feather ESP pins rather then my random pins?

ladyada: i'd be more amenable to it...but really the board definition should set the I2C pins, not the user....

marcelstoer: I do see your point. However, there are many situations where that information isn't available (missing or incorrect). How about going with the [initial idea](https://github.com/adafruit/RTClib/issues/129#issue-536043999) and just adding two lines of commented code?

ladyada: i think if you make it clear that it is only required for special boards it could be OK 

really i do not like it at all, this system of setting pins is completely different than all other Arduino boards and has only caused support problems and confusion, so you have to convince me.

marcelstoer: Feel free to close this PR; I don't mind. It was born out of frustration when I first used the lib with an example on a generic ESP8266. I wondered how I was supposed to set the correct pins.

ladyada: thanks - i think best thing to do is have the board maker you use define the I2C pins like every single other Arduino compatible ...this neglect on their part has caused a lot of problems :(

PR #133, opened 2020-02-19
Move the rtc.adjust to outside the if-statement

DanielFrykman:
If the rtc.adjust is inside the the if (! rtc.isrunning()) it will not run unless rtc.isrunning is not running, but when rtc.isrunning is not running it can't run rtc.adjust...
I have with success implemented this changes to my own clone of the library.

Thank you for creating a pull request to contribute to Adafruit's GitHub code!
Before you open the request please review the following guidelines and tips to
help it be more easily integrated:

- **Describe the scope of your change--i.e. what the change does and what parts
  of the code were modified.**  This will help us understand any risks of integrating
  the code.

- **Describe any known limitations with your change.**  For example if the change
  doesn't apply to a supported platform of the library please mention it.

- **Please run any tests or examples that can exercise your modified code.**  We
  strive to not break users of the code and running tests/examples helps with this
  process.

Thank you again for contributing!  We will try to test and integrate the change
as soon as we can, but be aware we have many GitHub repositories to manage and
can't immediately respond to every request.  There is no need to bump or check in
on a pull request (it will clutter the discussion of the request).

Also don't be worried if the request is closed or not integrated--sometimes the
priorities of Adafruit's GitHub code (education, ease of use) might not match the
priorities of the pull request.  Don't fret, the open source community thrives on
forks and GitHub makes it easy to keep your changes in a forked repo.

After reviewing the guidelines above you can delete this text from the pull request.

### DIFF
--- a/examples/ds1307/ds1307.ino
+++ b/examples/ds1307/ds1307.ino
@@ -17,11 +17,11 @@ void setup () {
   if (! rtc.isrunning()) {
     Serial.println("RTC is NOT running!");
   }
-    // following line sets the RTC to the date & time this sketch was compiled
-    // rtc.adjust(DateTime(F(__DATE__), F(__TIME__)));
-    // This line sets the RTC with an explicit date & time, for example to set
-    // January 21, 2014 at 3am you would call:
-    // rtc.adjust(DateTime(2014, 1, 21, 3, 0, 0));
+  // following line sets the RTC to the date & time this sketch was compiled
+  // rtc.adjust(DateTime(F(__DATE__), F(__TIME__)));
+  // This line sets the RTC with an explicit date & time, for example to set
+  // January 21, 2014 at 3am you would call:
+  // rtc.adjust(DateTime(2014, 1, 21, 3, 0, 0));
 }
 
 void loop () {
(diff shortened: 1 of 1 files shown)


examples/ds1307/ds1307.ino
```diff
@@ -16,12 +16,12 @@ void setup () {
 
   if (! rtc.isrunning()) {
     Serial.println("RTC is NOT running!");
+  }
     // following line sets the RTC to the date & time this sketch was compiled
```
edgar-bonet: Then these comments should be indented one level less.

DanielFrykman: Agreed

drak7:
Typically the RTC only needs a time set if it lost power and isn't running, but I can see why you'd want to set one in other cases.
Can you do this for the other examples as well (ds3231, pcf8523)?

DanielFrykman:
True. In my case it was because I live in a different time zone then the rtc was set to. 
So you need an easy way of setting the time and I suppose that just uncommenting is the simplest way. 

I will look into the other examples in the upcoming days.

colindgrant:
> Typically the RTC only needs a time set if it lost power and isn't running, but I can see why you'd want to set one in other cases.
> Can you do this for the other examples as well (ds3231, pcf8523)?

@drak7 I just filed #156 to cover the other examples. I'll give it a go and submit a PR for your consideration.

Issue #140, opened 2020-04-02
12 Hour Format Support


jadonmmiller:
I'm working on a moderately sized project, and I'm considering modifying the DateTime class to support the twelve-hour format.

I'm thinking of storing the time internally in the 24hr format as it is now, but add an additional constructor with something like an *isPM* argument. (0 = AM, 1 = PM - true and false should also work.)
```C++
DateTime (uint16_t year, uint8_t month, uint8_t day,
                     uint8_t hour = 0, uint8_t isPM = 0,
                     uint8_t min = 0, uint8_t sec = 0);
```
I would also need to add another hour function that would return the hour with a range of 1-12, maybe something like this.
```C++
uint8_t twelveHour() const 
```
As well as an AM/PM function:
```C++
uint8_t isPM() const
```

I'm just wondering if there's any interest in this functionality, and if I do start working on it, any warnings or suggestions. Thanks, and feel free to tell me any ideas you have!

drak7:
Closing this since we have #141

PR #141, opened 2020-04-03
Add DateTime Support for 12 Hour Format

jadonmmiller:
This is the change discussed in issue #140. There I asked if there was interest, and while I didn't get a response, I coded 12-hour support for my own use. I'd be honored to have this implemented, but if you don't feel it should be done, or want modifications, I understand.

### Goal
This adds support for the twelve-hour AM/PM time format natively to the DateTime library. Thus, it can be used with any supported RTC, or anywhere else the library is used.

### Scope
This doesn't change any of the existing code, so it should be completely backwards compatible. In fact, this implementation stores the time in 24hour format like RTClib always has, so you can switch between 12 and 14 Hour format at will. (Initialize in 24 hr, then display the time in 12 hr, or vice versa.

### Changes
This adds four new functions:
- DateTime (year, month, day, hour, isPM, minute, second) - (Constructor) This converts the time to 24 hr, then stores it.
- to12HrString - The same as the toString function, but adds an 'aa' tag that returns "AM" or "PM" and changes the 'hh' tag to return the hours reformatted to 12-hour time.
- twelveHour - Same as hour() function, but returns the time mapped to fit the 12-hour format. The original function is unchanged.
- isPM - Returns true if the stored time is PM, and false if it's AM.

### Testing
I tested the code with the sketch below, which prints over serial the minutes in a day in twelve and twenty-four hour formats. It displays the times using toString(), to12hrString(), and the getter functions. (hour(), minute(), second(), etc.)

```C++
#include <RTClib.h>

DateTime time(2020, 4, 3, 0, 0, 0);

void setup() {
  Serial.begin(115200);

  //Repeat to increment the time
  for (int i = 0; i < 1440; i++) {

    Serial.println("Testing toString Functions:");

    char buf1[] = "hh:mm";
    Serial.print("24 Hour Time: ");
    Serial.println(time.toString(buf1));

    char buf2[] = "hh:mm aa";
    Serial.print("12 Hour Time: ");
    Serial.println(time.to12hrString(buf2));

    char buf3[] = "YYMMDD-hh:mm:ss";
    Serial.print("24 Hour Time with Date: ");
    Serial.println(time.toString(buf3));

    char buf4[] = "YYMMDD-hh:mm:ss aa";
    Serial.print("12 Hour Time with Date: ");
    Serial.println(time.to12hrString(buf4));


    Serial.println();
    Serial.println("Testing Getter Functions:");
    Serial.println("--- 24 Hour ---");
    Serial.print("Hour: ");
    Serial.println(time.hour());
    Serial.print("Minute: ");
    Serial.println(time.minute());
    Serial.print("Seconds: ");
    Serial.println(time.second());

    Serial.println();
    Serial.println("--- 12 Hour ---");
    Serial.print("Hour: ");
    Serial.println(time.twelveHour());
    Serial.print("Minute: ");
    Serial.println(time.minute());
    Serial.print("Seconds: ");
    Serial.println(time.second());
    Serial.print("AM/PM: ");
    if (time.isPM()) {
      Serial.println("PM");
    }
    else {
      Serial.println("AM");
    }

    //Add spacing between printouts
    Serial.println();
    Serial.println();
    Serial.println();
    Serial.println();

    //increment the time by a minute
    time = time + TimeSpan(0, 0, 1, 0);
  }
}

void loop() {
  //We don't need to do anything in the loop
}
```

Please feel free to ask questions, give suggestions, or make changes. Thanks!

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -513,21 +513,6 @@ uint8_t DateTime::twelveHour() const {
 
 /**************************************************************************/
 /*!
-   @brief  Return whether the current time is PM
-   @return uint8_t False = AM, True = PM
-*/
-/**************************************************************************/
-uint8_t DateTime::isPM() const {
-    if (hh < 12) { //morning
-        return false;
-    }
-    else if (hh >= 12) { //evening
-        return true;
-    }
-}
-
-/**************************************************************************/
-/*!
     @brief  Return the day of the week for this object, from 0-6.
     @return Day of week 0-6 starting with Sunday, e.g. Sunday = 0, Saturday = 6
 */
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -342,17 +342,46 @@ DateTime::DateTime (const __FlashStringHelper* date, const __FlashStringHelper* 
     @param buffer: array of char for holding the format description and the formatted DateTime. 
                    Before calling this method, the buffer should be initialized by the user with 
                    a format string, e.g. "YYYY-MM-DD hh:mm:ss". The method will overwrite 
-                   the buffer with the formatted date and/or time.
+                   the buffer with the formatted date and/or time. Use the "AP" tag to
+                   return "AM" or "PM" and 12-Hour time.
     @return a pointer to the provided buffer. This is returned for convenience, 
             in order to enable idioms such as Serial.println(now.toString(buffer));
 */
 /**************************************************************************/
 
-char* DateTime::toString(char* buffer){
-		for(int i=0;i<strlen(buffer)-1;i++){
+char* DateTime::toString(char* buffer){        
+    char* apTag = strstr(buffer, "ap");
+    char* APTag = strstr(buffer, "AP");
+    uint8_t hourReformatted, isPM;
+    if(apTag != nullptr || APTag != nullptr) { //12 Hour Mode        
+        if(hh == 0) { //midnight
+            isPM = false;
+            hourReformatted = 12;
+        }
+        else if(hh == 12) { //noon
+            isPM = true;
+            hourReformatted = 12;
+        }
+        else if(hh < 12) { //morning
+            isPM = false;
+            hourReformatted = hh;
+        }
+        else { //1 o'clock or after
+            isPM = true;
+            hourReformatted = hh - 12;
+        }
+    }
+
+	for(int i=0;i<strlen(buffer)-1;i++){
 		if(buffer[i] == 'h' && buffer[i+1] == 'h'){
-			buffer[i] = '0'+hh/10;
-			buffer[i+1] = '0'+hh%10;
+            if (apTag == nullptr && APTag == nullptr) { //24 Hour Mode
+                buffer[i] = '0' + hh / 10;
+                buffer[i + 1] = '0' + hh % 10;
+            }
+            else { //12 Hour Mode
+                buffer[i] = '0' + hourReformatted / 10;
+                buffer[i + 1] = '0' + hourReformatted % 10;
+            }
 		}
 		if(buffer[i] == 'm' && buffer[i+1] == 'm'){
 			buffer[i] = '0'+mm/10;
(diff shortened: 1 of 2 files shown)


edgar-bonet:
> to12HrString - The same as the toString function, but adds an 'aa' tag that returns "AM" or "PM" and changes the 'hh' tag to return the hours reformatted to 12-hour time.

I think it would make more sense to add 12-hour capability to the existing `toString()` method. Otherwise we end up with a lot of duplicate code.

I would rather use “AP” for the new tag, as this method is loosely modeled on [Qt's `QDateTime::toString()`][qt], which [uses the “AP” tag][tag]. Also note that Qt's “hh” tag displays 12-hour time if the “AP” tag is used, and 24-hour time otherwise.

I am no fan of the other new methods, as I think those 12h ↔ 24h conversions should be the responsibility of the user code. But I do think 12h support deserves a place in `DateTime::toString()`. This is just a personal opinion though, and I am only a very minor contributor here.

[qt]: https://doc.qt.io/qt-5/qdatetime.html#toString
[tag]: https://doc.qt.io/qt-5/qtime.html#toString

RTClib.cpp
```diff
@@ -361,6 +399,133 @@ char* DateTime::toString(char* buffer){
...
+   @return uint8_t False = AM, True = PM
+*/
+/**************************************************************************/
+uint8_t DateTime::isPM() const {
+    if (hh < 12) { //morning
+        return false;
+    }
+    else if (hh >= 12) { //evening
+        return true;
+    }
```
edgar-bonet, line 526:
```suggestion
    return hh >= 12;
```

Not only is this simpler, it should also make Travis happy.

jadonmmiller: That, friend, is the beauty of open source! You can take my bulky code and make it super-efficient!

Now that we can do that function inline, should I move it to the .h file?

Also, could you please explain to me why a Travis build would fail? Thanks - I'm new to this!

jadonmmiller:
> I think it would make more sense to add 12-hour capability to the existing `toString()` method. Otherwise we end up with a lot of duplicate code.
> 
> I would rather use “AP” for the new tag, as this method is loosely modeled on [Qt's `QDateTime::toString()`](https://doc.qt.io/qt-5/qdatetime.html#toString), which [uses the “AP” tag](https://doc.qt.io/qt-5/qtime.html#toString). Also note that Qt's “hh” tag displays 12-hour time if the “AP” tag is used, and 24-hour time otherwise.

👍 Great! I'm a fan of doing things right! I wasn't aware of the Qt function.

I didn't previously combine the two functions because I hadn't thought of looking for the AP tag to determine the time format. Would that be best implemented with a simple scan-a-head *for* loop looking for the AP tag?

Maybe this is feature creep, but it would be easy to implement a lowercase version, as well as a single letter version. ('A' could return 'A' or 'P' instead of the full "AM" or "PM") My only concern would be using the function for things other than simple time, like in the [toString example sketch.](https://github.com/adafruit/RTClib/blob/master/examples/toString/toString.ino) ("Today is DDD, MMM DD YYYY"

> I am no fan of the other new methods, as I think those 12h ↔ 24h conversions should be the responsibility of the user code. 

Are you referring to the "getter" functions? (isPM() and twelveHour()) They give the user code much more flexibility as to making time-based decisions. If you want though, we could add an optional argument to the hour() function that would set it to 12hr mode or even eliminate the functions if the community agrees; it's only my opinion!

Thanks!
Jadon

RTClib.cpp
edgar-bonet: > Now that we can do that function inline, should I move it to the .h file?

I would think so. This .h file is already inlining quite a few trivial functions, so inlining this one would be consistent with the current style.

> Also, could you please explain to me why a Travis build would fail?

If you click on “Show all checks” next to “All checks have failed”, you will see a mention of Travis, with a “Details” link to the right. Follow the link, then the first link on the next page (“The build”). You will see the build output messages. Search for “error” and you will see:

```text
.../RTClib.cpp: In member function 'uint8_t DateTime::isPM() const':
.../RTClib.cpp:527:1: error: control reaches end of non-void function [-Werror=return-type]
 }
 ^
cc1plus: some warnings being treated as errors
```

What happens is that the compiler saw this structure:

```c++
uint8_t DateTime::isPM() const {
    if (some_condition)
        return false;
    else if (some_other_condition)
        return true;
    // And what if both conditions are false?
    // What is the function then supposed to return?
}
```

Clearly the compiler is not being very clever, as both conditions cannot be simultaneously false. This error only happens with the ESP32 build chain. A better optimizing compiler would likely optimize away the second `if` test.

However, it never hurts to make the logic of the code obvious at the syntax level. If you write:

```c++
if (some_condition)
    return false;
else
    return true;
```

it is immediately obvious, even to the dumbest compiler, that only one condition has to be tested, and that a value is always returned. Some style guides even recommend to drop the `else` keyword in order to make the last `return` unconditional.

edgar-bonet:
> Would [switching the “hh” tag to 12h] be best implemented with a simple scan-a-head for loop looking for the AP tag?

Or just call the standard library function `strstr()`, which does exactly that.

> but it would be easy to implement a lowercase version, as well as a single letter version.

Lowercase, why not. But would anyone want a single-letter version? Note that Qt uses “A” and “a” as synonyms for “AP” and “ap”. They don't provide single-letter output. This can't work with this API that uses a single buffer for format and output. Also, as you already noted, a single letter code has a high risk of collision.

> [...] or even eliminate the functions if the community agrees [...]

More community feedback would be good. At the end it's Adafruit's call.

RTClib.cpp
jadonmmiller: > I would think so. This .h file is already inlining quite a few trivial functions, so inlining this one would be consistent with the current style.

My thoughts exactly.

Thanks for explaining Travis. I only made it about half-way through the link-chain! 😄

jadonmmiller:
> Or just call the standard library function `strstr()`, which does exactly that.

Great idea!

> But would anyone want a single-letter version?

I don't know. I'd probably be best to just stay with an upper and lowercase two-letter version.

Thanks for your help through all this, Edgar! This is one of my first open source contributions, and I love the community feel. I'll commit the new changes by tonight. Thanks!

drak7:
I agree that it's good to keep the internals 24-hour, it keeps things simple and saves some space. I don't think another constructor is necessary, let's have the user deal with the conversion if they need to. The helpers are lightweight enough to keep and the toString piece is useful.

RTClib.cpp
```diff
@@ -304,17 +342,46 @@ DateTime::DateTime (const __FlashStringHelper* date, const __FlashStringHelper*
...
     @return a pointer to the provided buffer. This is returned for convenience, 
             in order to enable idioms such as Serial.println(now.toString(buffer));
 */
 /**************************************************************************/
 
-char* DateTime::toString(char* buffer){
-		for(int i=0;i<strlen(buffer)-1;i++){
+char* DateTime::toString(char* buffer){        
+    char* apTag = strstr(buffer, "ap");
+    char* APTag = strstr(buffer, "AP");
```
drak7, line 354: Since the actual location of the tag isn't used you could set a single flag here instead of checking both apTag and APTag later on. Something like `uint8_t apTag = ((strstr(buffer, "ap") != nullptr) || (strstr(buffer, "AP") != nullptr));`

jadonmmiller:
> 
> 
> I agree that it's good to keep the internals 24-hour, it keeps things simple and saves some space. I don't think another constructor is necessary, let's have the user deal with the conversion if they need to. The helpers are lightweight enough to keep and the toString piece is useful.

Your wish is my command, O great one! Seriously though, thanks for your suggestions and all the work you put into managing this code! I committed your apTag suggestion and eliminated the 12-hour constructor.

Thanks for your input!
- Jadon

PR #150, opened 2020-04-20
actionified, formatted and doxy'd

siddacious:


.github/workflows/githubci.yml
```diff
@@ -0,0 +1,32 @@
...
+      run: python3 ci/build_platform.py main_platforms
+
+    - name: clang
+      run: python3 ci/run-clang-format.py -e "ci/*" -e "bin/*" -r . 
+
+    - name: doxygen
+      env:
+        GH_REPO_TOKEN: ${{ secrets.GH_REPO_TOKEN }}
+        PRETTYNAME : "RTClib"
+      run: bash ci/doxy_gen_and_deploy.sh
```
edgar-bonet, line 32: This script expects to be run in a Travis build environment, and makes use of some Travis-specific environment variables. See for example commit fae1b6904b198a92e5133d159885cd801ee2ccaf: the author is incorrectly set to “travis@travis-ci.org”, and the commit message is incomplete. The script also fails to check whether the commit lands on master, thus if someone works on a feature branch without using a fork, the Doxygen docs for the feature branch will be published and replace the master docs.

README.md
```diff
@@ -47,3 +41,29 @@ STM32F2            |             |             |     X       |
   * ATtiny85 @ 8MHz : Adafruit Gemma, Arduino Gemma, Adafruit Trinket 3V
 
 <!-- END COMPATIBILITY TABLE -->
+Adafruit invests time and resources providing this open source code, please support Adafruit and open-source hardware by purchasing products from Adafruit!
+
+# Dependencies
+ * [TinyWireM](https://github.com/adafruit/TinyWireM)
```
edgar-bonet, line 47: Only on the ATtiny85 does RTClib depend on TinyWireM.

README.md
```diff
@@ -47,3 +41,29 @@ STM32F2            |             |             |     X       |
...
+ * [OneWire](https://github.com/PaulStoffregen/OneWire)
```
edgar-bonet, line 48: I fail to see any relationship between RTClib and OneWire.

PR #154, opened 2020-04-23
Fix examples to start Serial before checking for USB connection.

### LINKED ISSUE #137: Error in pcf8523 example
Hello. I have just the rfc8523.ino example on an Adafruit RTC-Logger Feather board (#2922) and Programino IDE. 
However I quickly found that it would compile and load but not run, but instead locks up without anything apparent happening on the board.

I found that there is a problem at the start of this example program.

void setup () {

  while (!Serial) {
    delay(1);  // for Leonardo/Micro/Zero
  }

  Serial.begin(57600);
  if (! rtc.begin()) {
    while (1);
  }

No doubt you see it straight away. It is waiting for the Serial to be ready before initialising it.
I have looked at a couple of the other examples, but they do not have the same issue. I have not looked at them all.
Cheers,

colindgrant:
Prompted by RTClib issue #137 for PCF8523.ino, but upon inspection, all of the RTClib examples that waited for a native USB connection were checking Serial status before it was actually started.

Further, some examples exempted ESP8266 from the Serial check, others did not have a check at all.

This change standardizes the Serial opening steps across all RTClib examples.

Tested on:
Feather M4 Express w/DS1307
Feather M4 Express w/PCF8523
Feather nRF52840 Sense w/DS1307
Feather nRF52840 Sense w/PCF8523

(in the case of the DS1307, 5v power was supplied, with SDA/SCL pulled up to Feather's 3v supply)

Unfortunately, I do not have an ESP8266 or DS3231 to test with. However, all examples connected to the Arduino IDE built-in Serial monitor without issue.

Useful reference: https://www.arduino.cc/reference/en/language/functions/communication/serial/ifserial

Resolves #137

Personal note - I'm new to this community but excited to contribute. I appreciate your review and welcome your feedback!

### DIFF
--- a/examples/datecalc/datecalc.ino
+++ b/examples/datecalc/datecalc.ino
@@ -43,11 +43,11 @@ void showTimeSpan(const char* txt, const TimeSpan& ts) {
 }
 
 void setup () {
+    Serial.begin(57600);
 
 #ifndef ESP8266
-  while (!Serial); // for Leonardo/Micro/Zero
+    while (!Serial); // wait for serial port to connect. Needed for native USB
 #endif
-    Serial.begin(57600);
 
     DateTime dt0 (0, 1, 1, 0, 0, 0);
     showDate("dt0", dt0);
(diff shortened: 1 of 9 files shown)


drak7:
Welcome and thanks for cleaning those up! I tested the DS3231 on an ESP8266 and it looks good.

Issue #88, opened 2018-01-09
Working correction for issue #81 and added function as proposed in issue #86


eriktheV-king:
The corrected .cpp file adding the readSqwPinMode() function in the DS3231 part of the .cpp file as proposed in issue #81 bugged when I tried to compile.

I managed to get it working as follows:
https://github.com/eriktheV-king/RTClib-from-Adafruit/blob/master/RTClib.cpp
-----------------------------------------------------------------------------------------
Ds3231SqwPinMode RTC_DS3231::readSqwPinMode() {
int mode;

Wire.beginTransmission(DS3231_ADDRESS);
Wire._I2C_WRITE(DS3231_CONTROL);
Wire.endTransmission();

Wire.requestFrom((uint8_t)DS3231_ADDRESS, (uint8_t)1);
mode = Wire._I2C_READ();

//mode &= 0x93;//bug due to using ds1307 read mask
mode &= 0x1C;
if(mode == 0x04)
mode = DS3231_OFF;
return static_cast<Ds3231SqwPinMode>(mode); // instead of "return static_cast(mode);"
}
-----------------------------------------------------------------------------------------

In the mean time, I also added the code for the .isrunning() function for DS3231 as proposed in issue # 86 and everything is now working perfectly well !

caternuson:
I think this was fixed by #210.

PR #155, opened 2020-04-23
Fix for DS3231 readSqwPinMode and century bit

### LINKED ISSUE #71: DS3231 not correctly reading date
According to the Maxim data sheet, the DS3231 has a century bit in BIT 7 of the 5th register. The current code does not mask this out, which means that the returned date is incorrect (2023 returned for 2016).  See the .c file line 471 which reads 
uint8_t d = bcd2bin(Wire._I2C_READ());

The same masking should be applied as is done in the seconds register
uint8_t d = bcd2bin(Wire._I2C_READ() & 0x7F);

Presumably the century flag should be incorporated somewhere else too ?? I didn't look at this since I dont need it at the moment.

This problem is visible if you hook up an Arduino to a Raspberry Pi on the same I2C bus, then have the RPi obtain time via NTP, then do a "hwclock -w" to write the current time to the RTC module.  Its not clear why the RPi sets the century flag given that we are in 2016, but that's a different issue !!
You can also demonstrate the same point by clearing the month bit with something like  
i2cset -f -y 1 0x68 0x05 0x11
However the RPi will reset the bit to 0x91 when the next "hwclock -w" is performed

I note that the same logic is used in the PCF8523 device, but I've not checked that data sheet. This should be checked against that data sheet too.

### LINKED ISSUE #81: DS3231 readSqwPinMode()
Thank you for opening an issue on an Adafruit Arduino library repository.  To
improve the speed of resolution please review the following guidelines and
common troubleshooting steps below before creating the issue:

- **Do not use GitHub issues for troubleshooting projects and issues.**  Instead use
  the forums at http://forums.adafruit.com to ask questions and troubleshoot why
  something isn't working as expected.  In many cases the problem is a common issue
  that you will more quickly receive help from the forum community.  GitHub issues
  are meant for known defects in the code.  If you don't know if there is a defect
  in the code then start with troubleshooting on the forum first.

- **If following a tutorial or guide be sure you didn't miss a step.** Carefully
  check all of the steps and commands to run have been followed.  Consult the
  forum if you're unsure or have questions about steps in a guide/tutorial.

- **For Arduino projects check these very common issues to ensure they don't apply**:

  - For uploading sketches or communicating with the board make sure you're using
    a **USB data cable** and **not** a **USB charge-only cable**.  It is sometimes
    very hard to tell the difference between a data and charge cable!  Try using the
    cable with other devices or swapping to another cable to confirm it is not
    the problem.

  - **Be sure you are supplying adequate power to the board.**  Check the specs of
    your board and plug in an external power supply.  In many cases just
    plugging a board into your computer is not enough to power it and other
    peripherals.

  - **Double check all soldering joints and connections.**  Flakey connections
    cause many mysterious problems.  See the [guide to excellent soldering](https://learn.adafruit.com/adafruit-guide-excellent-soldering/tools) for examples of good solder joints.

  - **Ensure you are using an official Arduino or Adafruit board.** We can't
    guarantee a clone board will have the same functionality and work as expected
    with this code and don't support them.

If you're sure this issue is a defect in the code and checked the steps above
please fill in the following fields to provide enough troubleshooting information.
You may delete the guideline and text above to just leave the following details:

- Arduino board:  **UNO**

- Arduino IDE version (found in Arduino -> About Arduino menu):  **1.8.1**

- List the steps to reproduce the problem below (if possible attach a sketch or
  copy the sketch code in too): **LIST REPRO STEPS BELOW**

https://forum.arduino.cc/index.php?topic=501028.0

There is a bug in the readSqwPinMode() function in the DS3231 portion of the library. It is incorrectly using the read mask from the DS1307 readSqwPinMode() function. If you modify the .cpp code in the library as below, the print out is correct.

Ds3231SqwPinMode RTC_DS3231::readSqwPinMode() {
  int mode;

  Wire.beginTransmission(DS3231_ADDRESS);
  Wire._I2C_WRITE(DS3231_CONTROL);
  Wire.endTransmission();
  
  Wire.requestFrom((uint8_t)DS3231_ADDRESS, (uint8_t)1);
  mode = Wire._I2C_READ();

  //mode &= 0x93;//bug due to using ds1307 read mask
  mode &= 0x1C;
  if(mode == 0x04)
	  mode = DS3231_OFF;
  return static_cast<Ds3231SqwPinMode>(mode);
}

drak7:
This fixes #71 by ignoring the century bit in the month register which could have caused problems if it overflowed or was set by something else (e.g. RPi) and then used with this library.

Fixes #81 and #88 by using the correct mask for readSqwPinMode() and includes an example to test it out.

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -1193,7 +1193,7 @@ DateTime RTC_DS3231::now() {
   uint8_t hh = bcd2bin(Wire._I2C_READ());
   Wire._I2C_READ();
   uint8_t d = bcd2bin(Wire._I2C_READ());
-  uint8_t m = bcd2bin(Wire._I2C_READ());
+  uint8_t m = bcd2bin(Wire._I2C_READ() & 0x7F);
   uint16_t y = bcd2bin(Wire._I2C_READ()) + 2000;
 
   return DateTime(y, m, d, hh, mm, ss);
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -1215,7 +1215,10 @@ Ds3231SqwPinMode RTC_DS3231::readSqwPinMode() {
   Wire.requestFrom((uint8_t)DS3231_ADDRESS, (uint8_t)1);
   mode = Wire._I2C_READ();
 
-  mode &= 0x93;
+  mode &= 0x1C;
+  if (mode & 0x04) {
+      mode = DS3231_OFF;
+  }
   return static_cast<Ds3231SqwPinMode>(mode);
 }
 
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -1217,7 +1217,7 @@ Ds3231SqwPinMode RTC_DS3231::readSqwPinMode() {
 
   mode &= 0x1C;
   if (mode & 0x04) {
-      mode = DS3231_OFF;
+    mode = DS3231_OFF;
   }
   return static_cast<Ds3231SqwPinMode>(mode);
 }
--- a/examples/ds3231SqwPin/ds3231SqwPin.ino
+++ b/examples/ds3231SqwPin/ds3231SqwPin.ino
@@ -1,0 +1,48 @@
+#include "RTClib.h"
+
+RTC_DS3231 rtc;
+
+int mode_index = 0;
+
+Ds3231SqwPinMode modes[] = { DS3231_OFF, DS3231_SquareWave1Hz,
+  DS3231_SquareWave1kHz, DS3231_SquareWave4kHz, DS3231_SquareWave8kHz };
+
+void print_mode() {
+  Ds3231SqwPinMode mode = rtc.readSqwPinMode();
+
+  Serial.print("Sqw Pin Mode: ");
+  switch(mode) {
+    case DS3231_OFF:              Serial.println("OFF");       break;
+    case DS3231_SquareWave1Hz:    Serial.println("1Hz");       break;
+    case DS3231_SquareWave1kHz:   Serial.println("1.024kHz");  break;
+    case DS3231_SquareWave4kHz:   Serial.println("4.096kHz");  break;
+    case DS3231_SquareWave8kHz:   Serial.println("8.192kHz");  break;
+    default:                      Serial.println("UNKNOWN");   break;
+  }
+}
+
+void setup () {
+  Serial.begin(57600);
+
+#ifndef ESP8266
+  while (!Serial); // for Leonardo/Micro/Zero
+#endif
+
+  if (! rtc.begin()) {
+    Serial.println("Couldn't find RTC");
+    while (1);
+  }
+
+  print_mode();
+}
+
+void loop () {
+  rtc.writeSqwPinMode(modes[mode_index++]);
+  print_mode();
+
+  if (mode_index > 4) {
+    mode_index = 0;
+  }
+
+  delay(5000);
+}


examples/ds3231SqwPin/ds3231SqwPin.ino
```diff
@@ -0,0 +1,48 @@
...
+void setup () {
+  Serial.begin(57600);
+
+#ifndef ESP8266
+  while (!Serial); // for Leonardo/Micro/Zero
+#endif
+
+  if (! rtc.begin()) {
+    Serial.println("Couldn't find RTC");
+    while (1);
```
colindgrant, line 33: In the course of #166 and #168, @edgar-bonet and I landed on this slight but important change to what should happen when the RTC is not reachable:

```
  if (! rtc.begin()) {
    Serial.flush();
    abort();
  }
```

I recommend that this nice new example follow that pattern unless there is a reason to deviate.

examples/ds3231SqwPin/ds3231SqwPin.ino
```diff
@@ -0,0 +1,48 @@
...
+    case DS3231_SquareWave8kHz:   Serial.println("8.192kHz");  break;
+    default:                      Serial.println("UNKNOWN");   break;
+  }
+}
+
+void setup () {
+  Serial.begin(57600);
+
+#ifndef ESP8266
+  while (!Serial); // for Leonardo/Micro/Zero
```
colindgrant, line 28: Trivial, but... There are many micros with USB Serial now, so while I was touching all the examples for #154, I decided to use the language from [the Arduino docs for if (Serial)](https://www.arduino.cc/reference/en/language/functions/communication/serial/ifserial):

`while (!Serial); // wait for serial port to connect. Needed for native USB`

In my opinion it clues new users into the significance of USB serial vs the significance of certain boards.

colindgrant:
Hi @drak7 I got a DS3231 Featherwing to play with so I decided to test this out.

So far I reproduced the original mode mask problem (described in the forum post), then pulled down your changes (`git fetch upstream pull/155/head:test155`), and verified that is fixed. Your example made it very easy, I added some comments there too, really just about standardization of the RTClib examples that's gone on since the parent of your example.

I'm planning to get out the datasheet for the two DS RTCs and try to understand what's going on under the hood as well. But so far this PR looks good to me. Let me know if there's some directed testing I can help with.

colindgrant:
Hi @drak, code review is a good learning tool for me, so I did a deeper dive to understand the changes here and with that knowledge I did some more testing too.

Here is what I think is going on:

The Century bit is the MSB (# 8) of the 'Month' register. `0x7F` is 1111111, which masks the Century bit so that when reading the month register, the Century bit is always 0.
I tested this by forcing the adjust() function to set the Century bit 1 with: `Wire._I2C_WRITE(bin2bcd(dt.month()) | 0x80);`.
With your change, this does not have an effect, the returned date is correct. Without your change, I get a unixtime() 2 years in the future, and an invalid month value. So your mask of `0x7F` works. I noticed the datasheet says the next two bits after the century bit should be 0's. So a more appropriate mask may be `0x1F` (which I also tested).

To read the square wave mode, we only need bit # 4-2 (RS2, RS1, INTCN) of the control register. `0x1c` is 11100, which masks all but those bits. So the bitwise AND with the contents of the control register to ignore other settings makes sense.

Square Wave Interrupt Control (INTCN) is bit # 2 of the control register. INTCN = 0 indicates square wave mode. `0x04` is 100, so bitwise AND with whatever the control register holds will be 1 when INTCN = 1, indicating square wave mode is NOT enabled. The if statement for determining `DS3231_OFF` makes sense.
What doesn't make sense to me is the enum value `DS3231_OFF = 0x01` which predated this change. It seems like an arbitrary value, and it's never actually used since readSqwPinMode and writeSqwPinMode both special-case it. If the enum were `DS3231_OFF = 0x04` instead, I think the if blocks in both functions could be removed.


If I got that right, then my humble opinion is that this code could be safely merged as-is. I'm interested in your feedback on the possible changes I noted above, but I think they're optional. If you'd still like to wait for Ladyada's review, that's fine too. This was a useful exercise for me regardless!

RTClib.cpp
```diff
@@ -1193,7 +1193,7 @@ DateTime RTC_DS3231::now() {
   uint8_t hh = bcd2bin(Wire._I2C_READ());
   Wire._I2C_READ();
   uint8_t d = bcd2bin(Wire._I2C_READ());
-  uint8_t m = bcd2bin(Wire._I2C_READ());
+  uint8_t m = bcd2bin(Wire._I2C_READ() & 0x7F);
```
colindgrant, line 1196: The datasheet's register address map shows bits 6 and 5 of the month register as always 0. So a more appropriate mask may be `0x1F`.
Reference: `Figure 1` here: https://datasheets.maximintegrated.com/en/ds/DS3231.pdf

RTClib.cpp
```diff
@@ -1215,7 +1215,10 @@ Ds3231SqwPinMode RTC_DS3231::readSqwPinMode() {
   Wire.requestFrom((uint8_t)DS3231_ADDRESS, (uint8_t)1);
   mode = Wire._I2C_READ();
 
-  mode &= 0x93;
+  mode &= 0x1C;
+  if (mode & 0x04) {
```
colindgrant, line 1219: If the header is updated with the following, then I think the special-case handling of DS3231_OFF in readSqwPinMode and writeSqwPinMode could be eliminated.

```
enum Ds3231SqwPinMode {
  DS3231_OFF = 0x01, 
  ...
```

misa-itk:
@drak7 I have tested your changes and SQW for **ds3231** finally works as **ds1307** example. Thank you for sharing the code. I hope this change will be soon approved and merged.

drak7:
See #210

Issue #132, opened 2020-01-22
API Compatability / example consistency


layzsmurf:
### RTClib 1.3.3

| Example  | DS1307 | DS3231 | PCF8523 |
| ------------- | ------------- |  ------------- | ------------- |
| toString  |     X   |   |   | 
| timestamp  |     X   |   |   | 
|  pcf8523 |        |   |  X  | 
|  ds3231 |      | x  |  | 
|  ds1307nvram |     X   |   |  | 
|  ds1307SqwPin |     X   |   |  | 
|  ds1307 |     X   |   |  | 
| datecalc  |       |   |  | 

### Observation 1: Some Functions with same functionality differ across classes

At this time it is not easy to replace one RTC with another as some tasks have different names across the classes, different return values  or some functionality the chips are not implemented. Once all classes have the same API calls - many examples could be merged into a simple example, where the user just needs to add/remove comments to match his RTC he is using.

| task | DS1307 | DS3231 | PCF8523 |
| ------------- | ------------- |  ------------- | ------------- |
| identify RTC powerloss |   uint8_t RTC_DS1307::isrunning  (  void  ) |   bool RTC_DS3231::lostPower  (  void  ) |  boolean RTC_PCF8523::initialized(void) |
| Read SQ Pin Mode | Ds1307SqwPinMode RTC_DS1307::readSqwPinMode()       |  Ds3231SqwPinMode RTC_DS3231::readSqwPinMode() |Pcf8523SqwPinMode RTC_PCF8523::readSqwPinMode() | 
| Change the SQW pin mode.  |    void RTC_DS1307::writeSqwPinMode(Ds1307SqwPinMode mode)   | void RTC_DS3231::writeSqwPinMode(Ds3231SqwPinMode mode)  | void RTC_PCF8523::writeSqwPinMode(Pcf8523SqwPinMode mode) | 
| Get the current temperature |       | float RTC_DS3231::getTemperature()  |  | 
|  calibrate RTC |       |   | void RTC_PCF8523::calibrate(Pcf8523OffsetMode mode, int8_t offset) 		 | 
| start RTC <sup>1</sup>  | boolean RTC_DS1307::begin (void ) 	| * | * |
| get current time  |   DateTime RTC_DS1307::now()    | *  | * | 
| adjust RTC time |    void RTC_DS1307::adjust(const DateTime &dt)   | *  | * | 

* = function in parameters and return value matches most left

[...]

PR #157, opened 2020-04-24
Consistent time setting statements across examples.

### LINKED ISSUE #156: Apply recent time setting improvements to additional examples.
Most examples include the same pair of time setting statements, which can set the RTC time to compile time, or to arbitrary user-chosen time.

#133 moves the time setting statements outside the if block that is entered when the RTC is not running, so that it time can also be re-set on a running RTC. This was only applied to ds1307.ino

#134 copies the time setting statements outside the if block that is entered when the RTC loses power, so that different things can happen on a freshly powered up RTC vs one that presumably already has time set. This was only applied to ds3231.ino.

The purpose of this issue is to extend those improvements to the rest of the RTClib examples (that is, pcf8523 and the additional ds1307 examples).

colindgrant:
Combines and extends #133 and #134 to additional examples, with some progress toward #132.

Each updated example sets the RTC to compile time after a power loss, as detected by the function specific to that RTC.
The same time-setting statements appear commented out later on for re-setting time on a running RTC, for whatever reason.

The PCF8523 initialized() state wasn't obvious to me. I found that when an unpowered PCF8523 is powered up it goes into standby mode, with Control_3 register set to a default value 0xE0, and in this state initialized() will return false. So it is closely associated with reporting whether power was lost. I made a brief comment to this effect in the member function.
The behavior is explained in section 8.5.1 of the datasheet https://www.nxp.com/docs/en/data-sheet/PCF8523.pdf

These changes tested on Feather M4 Express and Arduino Mega with DS1307 and PCF8523.
The DS3231 example has only cosmetic changes for consistency.

Fixes #156


examples/ds3231/ds3231.ino
```diff
@@ -81,4 +82,4 @@ void loop () {
 
     Serial.println();
     delay(3000);
-}
+}
```
edgar-bonet, line 85: For consistency, I would avoid removing the terminating newline character. Also note that all the files here adhere to the Unix end-of-line convention, which is to use LF as a line _terminator_ (not a line _separator_): every proper text file should end with LF.

colindgrant: Good catch, that was unintentional and I’ll fix it.

colindgrant: @edgar-bonet PR updated!

PR #158, opened 2020-04-25
Add new function RTC_PCF8523::lostPower()

### LINKED ISSUE #128: lostPower() for PCF8523
For the DS3231 you support reading the stop flag as per https://github.com/adafruit/RTClib/blob/04182cbbc10b242513c598ce397babb00609bf93/RTClib.cpp#L963. This request is to add a similar function for PCF8523.

Relevant data sheet section: https://www.nxp.com/docs/en/data-sheet/PCF8523.pdf#page=9 and chapter 8.6.1.1.

Your Python implementation for reference is here https://github.com/adafruit/Adafruit-uRTC/blob/master/urtc.py#L210 and the one for CircuitPython is here https://github.com/adafruit/Adafruit_CircuitPython_PCF8523/blob/master/adafruit_pcf8523.py#L75.

colindgrant:
This function mirrors the existing RTC_DS3231::lostPower()

The PCF8523 sets an Oscillator Stopped (OS) flag anytime its oscillator has stopped or been interrupted. This flag offers a way to verify clock integrity, and to know if power was lost.

To purposely set the OS flag, remove the battery and disconnect the RTC from external power supply for ~1 second or more.
When power is re-applied via battery or supply, the OS flag will remain set, even as the oscillator starts and stabilizes.

RTC_PCF8523::lostPower() returns true if the flag is set, and false only after it has been un-set, for example with RTC_PCF8523::Adjust()

This new function is somewhat similar to an existing function:

RTC_PCF8523::initialized() returns true when power management has been configured, for example with RTC_PCF8523::Adjust(), and false when power management configuration is found in 'standby mode', which is the default after power loss.

For many purposes they may be substitutes for each other, but note the inverted relationship of their return values.

These changes tested on Feather M4 Express and Arduino Mega with PCF8523-equipped Adalogger FeatherWing.

From the PCF8523 datasheet:
https://www.nxp.com/docs/en/data-sheet/PCF8523.pdf#page=20

8.6.1.1 Oscillator STOP flag
The OS flag is set whenever the oscillator is stopped. The flag remains
set until cleared by using the interface. When the oscillator is not running, then the OS flag
cannot be cleared. This method can be used to monitor the oscillator.
The oscillator may be stopped, for example, by grounding one of the oscillator pins, OSCI
or OSCO. The oscillator is also considered to be stopped during the time between
power-on and stable crystal resonance. This time may be in a range of 200 ms to 2 s,
depending on crystal type, temperature, and supply voltage. At power-on, the OS flag is
always set.

fixes #128


colindgrant:
Hi @drak7 et al,
This PR is really all about 38a17a9. It touches a file that is also in my previous PR #157 which hasn't been merged. So I branched from there attempting to avoid a 3-way merge and an untested code combination going into master, should both PRs be accepted. In hindsight I'm not sure if that was the right way to go. These are trivial changes but I'm trying to learn how to do things right! I'm happy to rearrange things to make it easier on your end, just let me know. Thanks! -Colin

marcelstoer:
I'm self-taught and pretty bad at reading electronics data sheets but this looks correct. Thank for the great explanations in the commit message and here; very professional.

drak7:
Looks like it will merge fine with #157 so no problem there.
The 200ms to 2s startup time stuck out to me, it might be worth noting in the library and the example. If the host runs adjust() before the oscillator stabilizes you could end up with OS = 1 even though the oscillator is running and you told it to clear that bit.

colindgrant:
@drak7 good point. That could be aggravating! I'll double check my understanding, think about how to word it, and push an update to this PR. Thanks.

drak7:
Just a note in the comments explaining it would be good, it probably won't affect most users but if someone is depending on the OS bit they could run into that issue.
Thanks!

colindgrant:
@drak7 _Please_ feel free to update/edit/move the comments I added just now.

I did try to trigger the scenario I described but I wasn't able to. Hopefully that means it's an unlikely one, but if somebody does hit it, they'll have a hint about it now.

PR #163, opened 2020-05-03
Add countdown timer support for PCF8523, and an example.

colindgrant:
With this change, the user may easily take advantage of additional timing capabilities of the PCF8523:

1) Enable a recurring 'Countdown Timer' with a range from microseconds to over 10 days.
The countdown timer has an interrupt enabled on the INT1 pin that is exposed as INT or SQW on Adafruit breakout boards.
At the end of every countdown, the INT pin is pulled low for an (optionally) adjustable period of time.
This allows a user to attach an interrupt that executes a piece of code at regular intervals, for sensor data collection or other uses.

2) Enable a 'Second Timer' (1Hz), requiring no parameters or configuration, for a fixed interval of 1 second.
This timer acts very much like a 1-second Countdown Timer, or the existing square wave timer with 'writeSqwPinMode(PCF8523_SquareWave1HZ)'.

The implementation of the Countdown Timer uses the PCF8523's 'Timer B' and there are couple reason for this choice:
- Timer B offers the advantage of programmable low pulse widths.
- Timer B can only function as a Countdown Timer, meaning enablement does not block any alternative uses.

Though not implemented at this time, the PCF8523's 'Timer A' may be used as a Watchdog Timer or Countdown Timer, with fixed width low pulses.
Given most MCUs have built in watchdog timer capability, and considering that Timer A and Timer B share the same INT pin exposed on Adafruit breakout boards, the additional value of Timer A appears to be limited. Still, steps were taken to allow future implementation of Timer A.

Enabling the Countdown or Second Timer causes any existing CLKOUT square wave, configured with writeSqwPinMode(), to halt.
Disabling the Countdown or Second Timer narrowly turns off just the chosen timer, preserving state flags for later reading, and preserving other configuration on the same registers.

A big-hammer function is provided to deconfigure all timers which comes in handy during development due to the battery backup on RTCs that prevents regular resets.

The included Arduino example exercises all functionality with and without interrupts, and attempts to educate a user on techniques and subtleties along the way.

Library source additions documented, tested, reviewed using Adafruit Doxyfile. Minor adjustments to existing documentation made for consistency across PCF8523 functions. Library source also formatted with clang-format.

Tested on PCF8523-equipped Adalogger FeatherWing with:
Arduino Mega 2560
Feather M4 Express
Feather nRF52840 Sense

The PCF8523 is available at low cost on a standalone breakout board, or on the Adalogger FeatherWing, so it seems worthy of additional library support. I needed timers for it personally and I have seen forum requests for timers as well. A previous effort to add this functionality seems to have stalled. If this PR is accepted, and there is interest in Timer A and Alarms, I can add those as well.

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -1086,8 +1086,8 @@ DateTime RTC_PCF8523::now() {
 
 /**************************************************************************/
 /*!
-    @brief  Read the mode of the SQW pin on the PCF8523
-    @return SQW pin mode as a Pcf8523SqwPinMode enum
+    @brief  Read the mode of the INT/SQW pin on the PCF8523
+    @return SQW pin mode as a #Pcf8523SqwPinMode enum
 */
 /**************************************************************************/
 Pcf8523SqwPinMode RTC_PCF8523::readSqwPinMode() {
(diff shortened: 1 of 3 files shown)


edgar-bonet:
The CI build failed on this pull request. Here is an excerpt of the error messages:

```text
SWITCHING TO arduino:avr:uno
[...]
	pcf8523Countdown.ino ❌ [...]
Global variables use 2075 bytes (101%) of dynamic memory,
leaving -27 bytes for local variables.
Maximum is 2048 bytes.
Not enough memory [...]
```

The new test sketch is taking too much RAM for the Uno. It appears most of this RAM is consumed by the embedded strings: I count about 1.3&nbsp;kB of string data. The easy fix is to wrap all those strings in [the `F()` macro][F], as in:

```c++
Serial.println(F("\nStarting PCF8523 Countdown Timer example."));
```

[F]: https://www.arduino.cc/reference/en/language/variables/utilities/progmem/#_the_f_macro

colindgrant:
> The easy fix is to wrap all those strings in [the `F()` macro](https://www.arduino.cc/reference/en/language/variables/utilities/progmem/#_the_f_macro)

Thanks @edgar-bonet! Just pushed an update and it's past the memory error.

The current error has to do with Doxygen reporting that I did not re-list params on an overload function. I used the @overload command along with a @brief explanation. The generated html seems pretty reasonable:
image

Should I re-list the @params even though they're already documented in the parent function? That seems error-prone but I can do it if that's typical. Or if you have an example of a library with overloads that meets doc requirements I can mimic that. Thanks.

edgar-bonet:
I would just want to point out that, with this feature, RTClib seems to be willing to play _extremely well_ with the timing functions provided by [avr-libc's `<time.h>`][time]. I bet most sketches using RTClib rely on `rtc.now()` to provide the current time, which is very slow, as it requires an I2C data exchange. In contrast, the libc function `time()` does little more than retrieve four bytes from RAM, which is extremely fast.

However, in order to use `time()` and the related standard C timing functions, you have to:

1. Initialize the system time with `set_system_time()`. I doesn't hurt that the epoch of `time_t` in avr-libc is the same as the one of `DateTime::secondstime()`, namely the start of year 2000.
2. Call `system_tick()` every second, which is expected to be typically done from an ISR. The “Second Timer” seems to be designed almost on purpose for this.

Here is a sketch showing how RTClib and `<time.h>` could play together. I could not test it for lack of the appropriate hardware:

```c++
#include <RTClib.h>
#include <time.h>

const uint8_t pin1Hz = 3;  // pin receiving the 1 Hz interrupt

RTC_PCF8523 rtc;

void setup() {
    Serial.begin(9600);

    // Initialize the RTC.
    rtc.begin();
    if (!rtc.initialized() || rtc.lostPower())

    // Initialize the system time from the RTC.
    set_system_time(rtc.now().secondstime());

    // Keep the time in sync using the 1 Hz output of the RTC.
    pinMode(pin1Hz, INPUT_PULLUP);
    rtc.enableSecondTimer();
    attachInterrupt(digitalPinToInterrupt(pin1Hz), system_tick, FALLING);
}

void loop() {
    time_t now = time(nullptr);
    Serial.println(isotime(gmtime(&now)));
    delay(1000);
}
```

Would it be worth mentioning this possibility in the documentation? Providing a test sketch along these lines? @colindgrant: would you like to test this idea on your Mega?

[time]: https://www.nongnu.org/avr-libc/user-manual/group__avr__time.html

colindgrant:
@edgar-bonet that's cool! Yes, I'll give it a try. I should have time in the next few days.
I'll reply here with my initial findings, or with questions, and then we can decide where to go from there next.

colindgrant:
@edgar-bonet your example works like a charm.
The only thing I had to add was:
`rtc.deconfigureAllTimers();`
That's because I had last run PCF8523Countdown.ino, which leaves the RTC with a .5 sec Countdown Timer. The Second Timer can be started even with a Countdown Timer already running, leading to extra calls to `system_tick`.

If your sketch were to be turned into an example, some light commenting might help for those unfamiliar with the time_t type and the time library. A check for AVR hardware might be good too. I noticed you chose input pin 3. Is that a typical choice for an AVR interrupt pin? I ask because I used pin 2 in the PCF8523Countdown.ino example, but I prefer to conform to any existing standards or norms that are out there in the world of examples.

How do you wish to proceed? I'm happy to test any code you create on my Mega/PCF8523 combo.

-Colin

edgar-bonet:
@colindgrant: My choice of pin 3 was somewhat random, not related to any standard practice. It may be better to switch to pin&nbsp;2 for consistency with pcf8523Countdown.ino.

I am not sure what the best course of action would be now. I will open an issue: that would probably be a better place for discussing the topic and maybe gathering extra opinions.

Issue #165, opened 2020-05-12
Document how to kick system_tick() from an RTC on AVR


edgar-bonet:
The canonical way to get the current time with RTClib seems to be by calling `rtc.now()`. This is a slow, blocking operation that requires an I2C data exchange. In contrast, [the avr-libc's &lt;time.h&gt;][avr] provides the standard function `time()`, which does little more than retrieve four bytes from RAM, and is thus extremely fast. However, in order to use `time()` and the related standard C timing functions, one has to:

1. Initialize the system time with `set_system_time()`. I doesn't hurt that the epoch of `time_t` in avr-libc is the same as the one of `DateTime::secondstime()`, namely the start of year 2000.
2. Call `system_tick()` every second, which is expected to be typically done from an ISR. The 1&nbsp;Hz “Second Timer” of the PCF8523 seems to be designed almost on purpose for this.

This approach is very efficient, as it only requires incrementing a global integer every second from an ISR, and reading it for getting the current time. Below is a test sketch demonstrating this approach:

```c++
#include <RTClib.h>
#include <time.h>

// Dummy implementation for non-AVR platforms.
#if !__AVR__
void set_system_time(time_t) {}
void system_tick() {}
#endif

const uint8_t pin1Hz = 2;  // pin receiving the 1 Hz interrupt

RTC_PCF8523 rtc;

void setup() {
    Serial.begin(9600);

    // Initialize the RTC.
    rtc.begin();
    if (!rtc.initialized() || rtc.lostPower())
        rtc.adjust(DateTime(F(__DATE__), F(__TIME__)));
    rtc.deconfigureAllTimers();  // just in case

    // Initialize the system time from the RTC.
    set_system_time(rtc.now().secondstime());

    // Keep the time in sync using the 1 Hz output of the RTC.
    pinMode(pin1Hz, INPUT_PULLUP);
    rtc.enableSecondTimer();
    attachInterrupt(digitalPinToInterrupt(pin1Hz), system_tick, FALLING);
}

void loop() {
    time_t now = time(nullptr);
    Serial.println(asctime(gmtime(&now)));
    delay(1000);
}
```

I now have the following questions:

1. Is it worth documenting this approach?
2. If so, what would be the right place(s) to do so? A .dox file that doxygen would include in the reference documentation? An example sketch? Both? A short note in the README pointing to either?
[...]

PR #166, opened 2020-05-12
Verify RTC communication in PCF8523 and DS1307 begin()

### LINKED ISSUE #64: ::begin() always return true
There is no way to check if RTC chip is not available because `begin ()` simply returns `true`.

On my sensor board I use workaround for DS3231 but similar code should work on other IC:

``` c
bool RTC_DS3231::begin(void) {
  Wire.begin();
  uint8_t x = read_i2c_register(DS3231_ADDRESS, 6); // 0x00..0x99 - Year in BCD
  return x != 0xFF; // got 0xFF if chip not present
}
```

colindgrant:
This change extends the logic already used for verifying comms with the DS3231 (ff898af) to the other two RTCs supported by this library: the PCF8523 and DS1307.

Fixes #64

I tested these controllers:
Feather M4 Express
Feather nRF52840 Sense
Arduino Mega

with these RTCs:
DS1307  DIY breakout board kit modified for pullup to 5v or 3v.
PCF8523 on RTC + SD Adalogger FeatherWing

Test Note: When `rtc.begin()` returns false, most examples print "Couldn't find RTC", and immediately go into an infinite `while(1)` loop. On the nRF52840 Sense, using native USB serial, the message never appears. Using a FTDI Friend on Serial1 *does print* the message as expected.

It's as if the message is buffered and there is no time to get it out through USB before entering the loop, and no ability to interrupt the loop to send it once inside.

These two workarounds cause the message to be printed as expected:
1) Call `Serial.flush()` prior to entering the while (1) loop
2) Add a brief delay inside the loop: `while (1) delay (1);`

I modified the PCF8523 examples to use flush(), but I left other examples with rtc.begin() alone for now. The newer, 3v nRF52 and the older, 5v DS1307 seem an unlikely combination, and I don't have a DS3231 to test. My plan is to bring this up with the adafruit/Adafruit_nRF52_Arduino maintainers, and see if there is an upstream fix or better workaround. I welcome any input the RTClib contributors have!

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -755,13 +755,16 @@ static uint8_t bin2bcd(uint8_t val) { return val + 6 * (val / 10); }
 
 /**************************************************************************/
 /*!
-    @brief  Startup for the DS1307
-    @return Always true
+    @brief  Start I2C for the DS1307 and test succesful connection
+    @return True if Wire can find DS1307 or false otherwise.
 */
 /**************************************************************************/
 boolean RTC_DS1307::begin(void) {
   Wire.begin();
-  return true;
+  Wire.beginTransmission(DS1307_ADDRESS);
+  if (Wire.endTransmission() == 0)
+    return true;
+  return false;
 }
 
 /**************************************************************************/
@@ -995,13 +998,16 @@ DateTime RTC_Micros::now() {
 
 /**************************************************************************/
 /*!
-    @brief  Start using the PCF8523
-    @return True
+    @brief  Start I2C for the PCF8523 and test succesful connection
+    @return True if Wire can find PCF8523 or false otherwise.
 */
 /**************************************************************************/
 boolean RTC_PCF8523::begin(void) {
   Wire.begin();
-  return true;
+  Wire.beginTransmission(PCF8523_ADDRESS);
+  if (Wire.endTransmission() == 0)
+    return true;
+  return false;
 }
 
 /**************************************************************************/
--- a/examples/pcf8523/pcf8523.ino
+++ b/examples/pcf8523/pcf8523.ino
@@ -14,6 +14,7 @@ void setup () {
 
   if (! rtc.begin()) {
     Serial.println("Couldn't find RTC");
+    Serial.flush(); // ensure above text prints with nRF52 native USB Serial
     while (1);
   }
 
--- a/examples/pcf8523Countdown/pcf8523Countdown.ino
+++ b/examples/pcf8523Countdown/pcf8523Countdown.ino
@@ -41,6 +41,7 @@ void setup () {
 
   if (! rtc.begin()) {
     Serial.println("Couldn't find RTC");
+    Serial.flush(); // ensure above text prints with nRF52 native USB Serial
     while (1);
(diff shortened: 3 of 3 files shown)


colindgrant:
@drak7 @edgar-bonet I got some feedback in the issue I opened with the nRF52840 core folks, linked above. I'm not convinced the diagnosis matches the symptom, but it's a bit beyond my level of experience. It seems significant that the issue affects Feather nRF52840 but not Feather M4.

I'm planning to dive into understanding it when I get time this weekend. But if you have thoughts or guidance on the appropriate action, I'd love to hear it. Not urgent, obviously. Thanks, Colin

colindgrant:
After a good bit of reading, I think I understand the issue with Serial output being unreliable before entering the infinite `while(1)` loop.

The nRF52840 uses the TinyUSB stack, which does not rely on interrupts. It waits in line for CPU time to flush out Serial data. The `while(1)` effectively blocks TinyUSB from getting ever getting to the front of the line. In https://github.com/adafruit/Adafruit_nRF52_Arduino/issues/505 and https://github.com/adafruit/Adafruit_TinyUSB_Arduino/issues/51, Ha Thatch recommends `while (1) yield()`. The reason `while (1) delay (1)` works is that delay also yields.

My testing shows that `abort()` has the same effect on a non-TinyUSB board, like the Mega. In some cases only a few characters are printed to screen. You can see this behavior if you combine this PR with #165. In other cases where you want immediate termination, `abort()` is the right choice over `while(1)` by itself.
Repro:
1) combine code from #165 (b85f1a6) with the latest tip of master (which now includes this PR)
2) start up *without* an RTC attached - the new RTC device communication tests should cause `begin()` to return false.
3) You will drop inside the `if (! rtc.begin())` loop, but the Serial message will not print as expected.

I'm planning to test Ha's recommendation on the boards and RTCs that I tested for this PR.

edgar-bonet:
This is because `abort()` disables interrupts before hanging the CPU. One should call `Serial.flush()` before `abort()`, which I forgot to do in #165. I will amend that pull request later and include some of your suggestions.

colindgrant:
> This is because `abort()` disables interrupts before hanging the CPU. One should call `Serial.flush()` before `abort()`

@edgar-bonet That makes sense, and I imagine flush before abort should work. E.g.:

```
  if (! rtc.begin()) {
    Serial.flush(); // allow above serial print to complete before terminating (looping)
    abort();
  }
```

Is that preferable to the method Ha Thatch recommended for using yield()? E.g.:

```
  if (! rtc.begin()) {
    while (1) {
      yield(); // go no further, but allow above serial print to complete
    }
  }
```

Either one solves the original issue of having an undefined `while(1)` type of termination, so maybe I am splitting hairs...

edgar-bonet:
I think that `Serial.flush(); abort();` does a better job at conveying the programmer's intent to the reader. But as you say, this may be splitting hairs.

colindgrant:
@edgar-bonet Ok, I agree. I’ll test that and submit a new PR to update all the examples to match. This was a useful learning experience, thanks!

PR #168, opened 2020-05-14
Add example on using interrupts for timekeeping

edgar-bonet:
This new example configures the RTC to provide interrupts at 1&nbsp;Hz, which are used for keeping the time in avr-libc. The standard C timing functions are then available: `time()`, `gmtime()`, `mktime()`, etc. See the comments in the sketch for a full explanation.

This example compiles only on AVR-based boards. Empty files named `.{uno,leonardo,mega2560}.test.only` have been added in order to prevent the CI script from attempting a build on other platforms.

A side note about infinite, side-effect-free loops: most of the provided examples contain this code snippet:

```c++
if (! rtc.begin()) {
  while (1);
}
```

While the intents of the infinite loop is quite clear (to hang the processor), this idiom turns out to be [undefined behavior in C++][ub] (see also [this blog post][blog] for a deeper discussion). I prefer to call `abort()`, which does exactly the same but is guaranteed to do it right. It also more clearly conveys the programmer's intent.

@colindgrant: would you mind testing on your hardware?

[ub]: https://en.cppreference.com/w/cpp/language/ub#Infinite_loop_without_side-effects
[blog]: https://blog.regehr.org/archives/161

### DIFF
--- a/examples/interrupts1Hz/interrupts1Hz.ino
+++ b/examples/interrupts1Hz/interrupts1Hz.ino
@@ -1,0 +1,100 @@
+/***********************************************************************
+
+    Combining RTClib with the avr-libc timing functions
+    ===================================================
+
+The standard way of getting the current time and date with RTClib is to
+call the now() method of the appropriate RTC class. This, however, is
+somewhat slow, as it involves communicating with the RTC through the I2C
+bus. An alternative, more lightweight method involves configuring the
+RTC to deliver one pulse per second to an interrupt pin, and counting
+the seconds in the interrupt handler. The timekeeping is then entirely
+handled in the Arduino, with no I2C communication with the RTC other
+than during the initialization phase.
+
+On AVR-based Arduinos (Uno, Nano, Micro, ...), the Arduino core library
+is built on top of avr-libc, which is an implementation of the standard
+C library for the AVR platform. This library provides the standard C
+functions for handling time:[1] time(), gmtime(), mktime(), etc. The
+time() function is normally used to retrieve the current time from the
+operating system, but since we have no operating system, the avr-libc
+provides its own non-standard functions for implementing a time source:
+
+ - set_system_time() initializes the library's idea of the current time
+ - system_tick() steps the system time by one second.
+
+This sketch demonstrates how to combine RTClib and avr-libc in order to
+handle the timekeeping entirely on the Arduino from an interrupt
+delivered by the RTC:
+
+ - RTClib is used to configure the RTC and retrieve the initial time
+ - avr-libc is used for regular timekeeping
+
+This sketch only works on AVR-based Arduinos, as it relies on
+non-standard functions provided by avr-libc.
+
+[1] https://www.nongnu.org/avr-libc/user-manual/group__avr__time.html
+
+***********************************************************************/
+
+#include <RTClib.h>
+#include <time.h>  // standard C timing functions
+
+// Pin receiving the one-pulse-per-second signal from the RTC.
+// This should be an interrupt-capable pin.
+const uint8_t pin1pps = 2;
+
+// We will use the PCF8523 RTC, which has the handy "Second Timer".
+// Other RTCs could be used with their "square wave" output configured
+// to 1 Hz.
+RTC_PCF8523 rtc;
+
+void setup() {
+
+  // Initialize the serial port.
+  Serial.begin(57600);
+  while (!Serial); // wait for serial port to connect. Needed for native USB
+
+  // Initialize the RTC.
+  if (! rtc.begin()) {
+    Serial.println("Couldn't find RTC");
+    Serial.flush();
+    abort();
+  }
+  if (!rtc.initialized() || rtc.lostPower()) {
+    rtc.adjust(DateTime(F(__DATE__), F(__TIME__)));
+  }
+  rtc.deconfigureAllTimers();  // undo previous configuration, if any
+
+  // Initialize the system time from the RTC time. Both avr-libc and
+  // DateTime::secondstime() use the start of year 2000 as their
+  // reference "epoch".
+  set_system_time(rtc.now().secondstime());
+
+  // Keep the time in sync using the one-pulse-per-second output of the
+  // RTC as an interrupt source and calling system_tick() from the
+  // interrupt service routine.
+  pinMode(pin1pps, INPUT_PULLUP);
+  rtc.enableSecondTimer();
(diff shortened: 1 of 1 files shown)


colindgrant:
@edgar-bonet I like it! Tomorrow I can test it on my Mega and Uno.

colindgrant:
Hi @edgar-bonet, I was able to test this today and it works like a charm. Since you are not able to observe your example running directly, I took some pictures for extra context.

I tested your example unmodified using the PCF8523-equipped Adalogger Featherwing with a Mega and an UNO:

fullsizeoutput_1ef1

fullsizeoutput_1ef2

The 'Show timestamp' option in the Arduino IDE helps demonstrate that the time keeping is as expected (obviously offset by the duration between compile time and first initializing the RTC time values):

interrupts_1Hz_PCF8523_Adalogger

I also tested with a DS1307 configured for 1 Hz square wave output. This just took a few modifications:

```
/***********************************************************************

    Combining RTClib with the avr-libc timing functions
    ===================================================

The standard way of getting the current time and date with RTClib is to
call the now() method of the appropriate RTC class. This, however, is
somewhat slow, as it involves communicating with the RTC through the I2C
bus. An alternative, more lightweight method involves configuring the
RTC to deliver one pulse per second to an interrupt pin, and counting
the seconds in the interrupt handler. The timekeeping is then entirely
handled in the Arduino, with no I2C communication with the RTC other
than during the initialization phase.

On AVR-based Arduinos (Uno, Nano, Micro, ...), the Arduino core library
is built on top of avr-libc, which is an implementation of the standard
C library for the AVR platform. This library provides the standard C
functions for handling time:[1] time(), gmtime(), mktime(), etc. The
time() function is normally used to retrieve the current time from the
operating system, but since we have no operating system, the avr-libc
provides its own non-standard functions for implementing a time source:

 - set_system_time() initializes the library's idea of the current time
 - system_tick() steps the system time by one second.

This sketch demonstrates how to combine RTClib and avr-libc in order to
handle the timekeeping entirely on the Arduino from an interrupt
delivered by the RTC:

 - RTClib is used to configure the RTC and retrieve the initial time
 - avr-libc is used for regular timekeeping

This sketch only works on AVR-based Arduinos, as it relies on
non-standard functions provided by avr-libc.

[1] https://www.nongnu.org/avr-libc/user-manual/group__avr__time.html

***********************************************************************/

#include <RTClib.h>
#include <time.h>  // standard C timing functions

// Pin receiving the one-pulse-per-second signal from the RTC.
// This should be an interrupt-capable pin.
const uint8_t pin1pps = 2;

// Other RTCs could be used with their "square wave" output configured
// to 1 Hz.
RTC_DS1307 rtc;

void setup() {

  // Initialize the serial port.
  Serial.begin(57600);
  while (!Serial); // wait for serial port to connect. Needed for native USB

  // Initialize the RTC.
  if (! rtc.begin()) {
    abort();
  }
  if (!rtc.isrunning()) {
  }
  rtc.writeSqwPinMode(DS1307_SquareWave1HZ);  // 1Hz pulse from SQW pin

  // Initialize the system time from the RTC time. Both avr-libc and
  // DateTime::secondstime() use the start of year 2000 as their
  // reference "epoch".
  set_system_time(rtc.now().secondstime());

  // Keep the time in sync using the one-pulse-per-second output of the
  // RTC as an interrupt source and calling system_tick() from the
  // interrupt service routine.
  pinMode(pin1pps, INPUT_PULLUP);
  attachInterrupt(digitalPinToInterrupt(pin1pps), system_tick, FALLING);
}

void loop() {

  // From here on, we only use the standard C timing functions.
  time_t now = time(nullptr);
  Serial.println(asctime(gmtime(&now)));
  delay(1000);
}

```

interrupts_1Hz_DC1307_1Hz_SQW

Feedbac:
It is a very clear, easy to follow example. Your comments and background information should help orient new users quickly!

I had to look up how the nullptr works in this context. You might consider adding a comment like "returns a new time_t object that represents the current time".

One thing I noticed is that other RTClib examples do not use underscores in their names. This may be a case where one is called for, but I wanted to point it out for you to decide.

Thanks for the information on side-effect-free loops. That was an interesting link. I'll submit a PR to fix the other examples.

Cheers,
Colin

colindgrant:
@edgar-bonet there is actually a problem using `abort();` here. The Serial message just before it will not be printed. Or at least it's not guaranteed to be printed. I ran into this on #165 as well, I'd like to continue discussion there.

edgar-bonet:
@colindgrant wrote:
> I was able to test this today and it works like a charm.

Thanks! And thanks for your nice comments!

> You might consider adding a comment like "returns a new time\_t object that represents the current time".

I was reluctant to document this because it's a very standard C function. Maybe it would be useful for the intended audience though, but then I have to also document `gmtime()` and `asctime()`.

> One thing I noticed is that other RTClib examples do not use underscores in their names.

Oh, right, I will rename it for consistency.

> Thanks for the information on side-effect-free loops. That was an interesting link. I'll submit a PR to fix the other examples.

I am not sure how it will work on non-AVR platforms. On most platforms, `abort()` is supposed to stop the program and return control to the operating system. What should it do on a microcontroller with no operating system? On AVR, the choice was made to have `abort()` disable interrupts and then hang the CPU in a tight loop. I do not know whether other MCU platforms make the same choice.

> there is actually a problem using `abort();` here. The Serial message just before it will not be printed.

My bad, I forgot to `Serial.flush();`. Will fix that.

edgar-bonet:
This is because `abort()` disables interrupts before hanging the CPU. My mistake was to forget to call `Serial.flush()` before abort. I will amend my PR later.

edgar-bonet:
I just updated the pull request with three commits:

1. Renamed interrupts\_1Hz → interrupts1Hz, for consistency with the naming of the other examples.
2. Added `Serial.flush();` right before `abort();` in order to ensure that the error message gets printed if the RTC is not found.
3. Documented the C timing functions being used. Although these are completely standard, the intended audience may not be familiar with them.

I checked that point 2 works as intended. In order to test it, I merged the current master (tag 1.7.0) in my local repo (did not push the merge). This brought the contents of pull request #166, which makes `rtc.begin()` return `false` if the RTC is not found.

Mat I squash-rebase the branch on top of master? Or would it better to keep its history?

siddacious:
I'm more of a release bot, but I don't mind a good squarsh and rebase

colindgrant:
@edgar-bonet, @siddacious:
I re-tested this on Mega and Uno, using the latest commit (526ccc2). Looks good and works as expected. 👍

PR #169, opened 2020-05-19
Ensure serial message printed and proper termination when RTC not found.

colindgrant:
3f379ab (PR #166) added connection verification for the PCF8523 and DS1307 matching that of the DS3231. For each RTC, if begin() returns false, the intended behavior is to print a message, "Couldn't find RTC" to an attached serial terminal and then terminate.

This change is a follow-on that corrects two issues which were not seen as long as begin() always returned true, across all RTClib examples:
1) The method of termination in place was an endless loop: `while(1)`
However, as pointed out in PR #168, the behavior of such a loop is undefined. (See PR for interesting reading!) The `abort()` command effectively starts an endless loop, but without any side effects. It also makes clear to new users what the program is doing. H/T to contributor @edgar-bonet for this advice.

2) The message text "Couldn't find RTC" was not guaranteed to actually print on screen before the processor entered the endless loop.
This has to do with the way serial data is actually buffered and transfered down the line. On some boards, serial data is processed using interrupts. On others, such as the Feather nRF52840 which uses the TinyUSB stack, interrupts are not used at all, and serial processing waits for the CPU to become available. Whatever the case, `Serial.flush()` allows serial processing to finish before terminating with the endless loop.

An alternate way of handling 1 and 2, recommended by the TinyUSB author, is to enter the endless loop immediately, but call `yield()` inside to instruct the processor to go ahead and handle any latent requests that come in. This also makes the loop have defined behavior. E.g.:

```
  if (! rtc.begin()) {
    while (1) {
      yield();  // delay(1) would also suffice
    }
  }
```

A fuller discussion of termination methods can be viewed in PR #166.

### DIFF
--- a/examples/ds1307/ds1307.ino
+++ b/examples/ds1307/ds1307.ino
@@ -14,7 +14,8 @@ void setup () {
 
   if (! rtc.begin()) {
     Serial.println("Couldn't find RTC");
-    while (1);
+    Serial.flush();
+    abort();
   }
 
   if (! rtc.isrunning()) {
--- a/examples/ds1307SqwPin/ds1307SqwPin.ino
+++ b/examples/ds1307SqwPin/ds1307SqwPin.ino
@@ -44,7 +44,8 @@ void setup () {
 
   if (! rtc.begin()) {
     Serial.println("Couldn't find RTC");
-    while (1);
+    Serial.flush();
+    abort();
   }
 
   print_mode();
--- a/examples/ds1307nvram/ds1307nvram.ino
+++ b/examples/ds1307nvram/ds1307nvram.ino
@@ -20,7 +20,11 @@ void setup () {
   while (!Serial); // wait for serial port to connect. Needed for native USB
 #endif
 
-  rtc.begin();
+  if (! rtc.begin()) {
+    Serial.println("Couldn't find RTC");
+    Serial.flush();
+    abort();
+  }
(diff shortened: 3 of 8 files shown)


colindgrant:
Pretty cool test and verification harness you guys have set up. Is there a way to put branches in my fork through it before submitting a PR?
I read this guide https://learn.adafruit.com/the-well-automated-arduino-library and that helped me check some things over locally, but it looks like this repo is using GitHub actions instead of Travis CI. Any tips or pointers?

ladyada:
it should 'just work' if you have Actions activated for your github account :)

colindgrant:
@drak7, @ladyada, thanks! I found the 'Desk of Ladyada GitHub Actions' video shortly after posting here. What perfect timing!

It turns out I needed to add `GH_REPO_TOKEN` as a "secret" to my fork of this repo in order to make the gh-pages push work. Now I'm looking for some way to set that globally so it applies to any Adafruit repo I fork now or later. If you have any hints, I'd appreciate them.

I used to manage continuous build/test/release for Perforce Software, so CI is near and dear to me. I'm thrilled to see this system you've set up for hardware libraries. Among the many advantages, it's a nice safety net for new contributors like me. 👍

ladyada:
https://github.blog/changelog/2020-05-14-organization-secrets/

PR #170, opened 2020-06-06
Add functions for DS3231 32kHz Output

AidenDawn:
This Pull Request adds functionality for the DS3231 to enable, disable and check the status of the optional 32kHz clock output which is available on some DS3231 modules. 

The new functions are included right at the bottom of RTClib.cpp and are based on the existing functions to enable and disable the internal Alarms

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -1525,9 +1525,9 @@ bool RTC_DS3231::alarmFired(uint8_t alarm_num) {
 /**************************************************************************/
 void RTC_DS3231::enable32K(void) {
   uint8_t status = read_i2c_register(DS3231_ADDRESS, DS3231_STATUSREG);
-  status |= (0x1 << 3);
+  status |= (0x1 << 0x03);
   write_i2c_register(DS3231_ADDRESS, DS3231_STATUSREG, status);
-  //Serial.println( read_i2c_register(DS3231_ADDRESS, DS3231_STATUSREG), BIN);
+  // Serial.println(read_i2c_register(DS3231_ADDRESS, DS3231_STATUSREG), BIN);
 }
 
 /**************************************************************************/
@@ -1537,9 +1537,9 @@ void RTC_DS3231::enable32K(void) {
 /**************************************************************************/
 void RTC_DS3231::disable32K(void) {
   uint8_t status = read_i2c_register(DS3231_ADDRESS, DS3231_STATUSREG);
-  status &= ~(0x1 << 3);
+  status &= ~(0x1 << 0x03);
   write_i2c_register(DS3231_ADDRESS, DS3231_STATUSREG, status);
-  //Serial.println( read_i2c_register(DS3231_ADDRESS, DS3231_STATUSREG), BIN);
+  // Serial.println(read_i2c_register(DS3231_ADDRESS, DS3231_STATUSREG), BIN);
 }
 
 /**************************************************************************/
@@ -1550,5 +1550,5 @@ void RTC_DS3231::disable32K(void) {
 /**************************************************************************/
 bool RTC_DS3231::status32K(void) {
   uint8_t status = read_i2c_register(DS3231_ADDRESS, DS3231_STATUSREG);
-  return (status >> (uint8_t) 0x03) & 0x1;
+  return (status >> 0x03) & 0x1;
 }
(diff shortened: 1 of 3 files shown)


colindgrant:
@AidenDawn Nice feature. I went ahead and tried this out on my DS3231 Featherwing paired with a Feather M4 Express, and also an Arduino Mega 2560. It all appears to work as advertised.

Here's an example that exercises the new member functions:

```
// 32K square wave pin on DS3231 RTC connected via I2C and Wire lib
#include "RTClib.h"

RTC_DS3231 rtc;

unsigned long startMillis = 0;
volatile unsigned int numPulses = 0;
const int interval = 1000;

// const int interruptPin = 2;  // Most Arduinos
const int interruptPin = 5;  // Adafruit Feather M0/M4/nRF52840

void setup () {
  Serial.begin(57600);

#ifndef ESP8266
  while (!Serial); // wait for serial port to connect. Needed for native USB
#endif

  pinMode(interruptPin, INPUT_PULLUP);

  if (! rtc.begin()) {
    Serial.flush();
    abort();
  }

  print32KStatus(); // initial state

  rtc.disable32K(); // set
  print32KStatus(); // verify

  rtc.enable32K();  // set
  print32KStatus(); // verify

  // wait out any possible startup effects, like oscillator stabilization
  delay (1000);

  Serial.println("Counting pulses for approximately 1 second...");
  Serial.flush();
  startMillis = millis();
  attachInterrupt(digitalPinToInterrupt(interruptPin), countPulses, FALLING);
}

void print32KStatus () {
  if (rtc.status32K()){
    Serial.println("32K is currently enabled");
  } else {
    Serial.println("32K is currently disabled");
  }
}

void countPulses () {
  numPulses++;
}

void loop () {
  if (millis() - startMillis >= interval) {
    detachInterrupt(digitalPinToInterrupt(interruptPin));
    Serial.print("Number of pulses: ");
    Serial.println(numPulses);
    Serial.flush();
    abort();
  }
}
```

Looks good overall! I just have a few pieces of feedback for you:
1) If the `RTC_DS3231::status32K()` function were instead named `RTC_DS3231::isEnabled32K()`, then code like this would be quite intuitive:

```
  if (rtc.isEnabled32K()){
    Serial.println("32K is currently enabled");
  }
```

2) It seems worth noting somewhere, perhaps in the enable32K doxygen comments, that the RTC starts up with 32K output enabled by default, and that a pullup resistor is required to use it.

3) For syntax highlighting in the Arduino IDE, put the new function names into keywords.txt at the root of this library.

I'm not one of the maintainers of this library, just a random user and contributor. So these are nothing more than suggestions. Hope this helps!

AidenDawn:
Thanks for the feedback, you're right on those points. I've included them and I've also modified your example to make use of the new function name

```
// 32K square wave pin on DS3231 RTC connected via I2C and Wire lib
#include "RTClib.h"

RTC_DS3231 rtc;

unsigned long startMillis = 0;
volatile unsigned int numPulses = 0;
const int interval = 1000;

// const int interruptPin = 2;  // Most Arduinos
const int interruptPin = 5;  // Adafruit Feather M0/M4/nRF52840

void setup () {
  Serial.begin(57600);

#ifndef ESP8266
  while (!Serial); // wait for serial port to connect. Needed for native USB
#endif

  pinMode(interruptPin, INPUT_PULLUP);

  if (! rtc.begin()) {
    Serial.flush();
    abort();
  }

  print32KStatus(); // initial state

  rtc.disable32K(); // set
  print32KStatus(); // verify

  rtc.enable32K();  // set
  print32KStatus(); // verify

  // wait out any possible startup effects, like oscillator stabilization
  delay (1000);

  Serial.println("Counting pulses for approximately 1 second...");
  Serial.flush();
  startMillis = millis();
  attachInterrupt(digitalPinToInterrupt(interruptPin), countPulses, FALLING);
}

void print32KStatus () {
  if (rtc.isEnabled32K()){
    Serial.println("32K is currently enabled");
  } else {
    Serial.println("32K is currently disabled");
  }
}

void countPulses () {
  numPulses++;
}

void loop () {
  if (millis() - startMillis >= interval) {
    detachInterrupt(digitalPinToInterrupt(interruptPin));
    Serial.print("Number of pulses: ");
    Serial.println(numPulses);
    Serial.flush();
    abort();
  }
}
```

PR #173, opened 2020-06-25
Added DateTime(char *iso8601date) constructor.

gounselor:
Please consider this small patch to your lib. 

It only creates an additional DateTime constructor which creates a DateTime object from a
ISO8601 formatted string. (e.g. 2020-06-25T15:29:37) 
See: https://en.wikipedia.org/wiki/ISO_8601

Limitations of this change are, that only years > 2000 will be correct, as only the yOffs is parsed,
as this library is missing a conv4d function. 

The code should not break anything as it is a complete new constructor, the only one 
which accepts a single const char *.

Best regards,

Michael

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -369,6 +369,36 @@ DateTime::DateTime(const __FlashStringHelper *date,
 
 /**************************************************************************/
 /*!
+    @brief  Constructor for creating a DateTime from an ISO8601 date string
+
+    This constructor expects its parameters to be a string in the 
+    https://en.wikipedia.org/wiki/ISO_8601 format, e.g:
+
+    "2020-06-25T15:29:37"
+
+	The year must be > 2000, as only the offset is considered.
+
+    Usage:
+
+    ```
+    DateTime dt("2020-06-25T15:29:37");
+    ```
+
+*/
+/**************************************************************************/
+DateTime::DateTime(const char *iso8601date) {
+  yOff = conv2d(iso8601date + 2);
+  m    = conv2d(iso8601date + 5);
+  d    = conv2d(iso8601date + 8);
+  hh   = conv2d(iso8601date + 11);
+  mm   = conv2d(iso8601date + 14);
+  ss   = conv2d(iso8601date + 17);
+}
+
+
+
+/**************************************************************************/
+/*!
     @brief  Check whether this DateTime is valid.
     @return true if valid, false if not.
 */
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -376,13 +376,16 @@ DateTime::DateTime(const __FlashStringHelper *date,
 
     "2020-06-25T15:29:37"
 
-        The year must be > 2000, as only the offset is considered.
-
     Usage:
 
     ```
     DateTime dt("2020-06-25T15:29:37");
     ```
+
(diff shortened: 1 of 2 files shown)


dherrada:
@gounselor Hey, this needs to be clanged for the CI to pass. Not sure if you're familiar with clang-format, but after installing it, for this specific case you'll need to do `clang-format -i RTClib.cpp`

gounselor:
> @gounselor Hey, this needs to be clanged for the CI to pass. Not sure if you're familiar with clang-format, but after installing it, for this specific case you'll need to do `clang-format -i RTClib.cpp`

Oh sorry, i wasn't even aware of that. I have now a new commit which is clang-formatted. 

How to proceed?

dherrada:
@gounselor So, the good news is that the clang did work. However, now you've got to fully document it. A great guide about doing that can be found here: https://learn.adafruit.com/the-well-automated-arduino-library/doxygen

This doesn't need too much, so to save you some time, I'll just tell you what you need to do.

On line 366 of RTClib.cpp put the following:
`@param iso8601date XXX`
Replace XXX with a description of what `iso8601date` is.

gounselor:
> @gounselor So, the good news is that the clang did work. However, now you've got to fully document it. A great guide about doing that can be found here: https://learn.adafruit.com/the-well-automated-arduino-library/doxygen
> 
> This doesn't need too much, so to save you some time, I'll just tell you what you need to do.
> 
> On line 366 of RTClib.cpp put the following:
> `@param iso8601date XXX`
> Replace XXX with a description of what the parameter `iso8601date` is.

Thanks, updated the documentation and just to be sure did run clang-format again. 
Will have a look at the guide. 

Best regards,

Michael

dherrada:
Great. I'll try to test this sometime soon. I doubt it'll be broken, but better safe than sorry.

dherrada:
Apparently I'm out of CR1220 batteries, so I've ordered some more and will test this PR when they get here

gounselor:
> Apparently I'm out of CR1220 batteries, so I've ordered some more and will test this PR when they get here

That's ok. I added a small change to have some sane values if the user supplied string is too short.
I hope that is ok.

RTClib.cpp
```diff
@@ -367,6 +367,40 @@ DateTime::DateTime(const __FlashStringHelper *date,
...
+
+    @param iso8601dateTime
+           A dateTime string in iso8601 format,
+           e.g. "2020-06-25T15:29:37".
+
+*/
+/**************************************************************************/
+DateTime::DateTime(const char *iso8601dateTime) {
+  char ref[] = "2000-01-01T00:00:00";
+  memcpy(ref, iso8601dateTime, strlen(ref));
```
edgar-bonet, line 395: The third argument should be `min(strlen(ref), strlen(iso8601dateTime))`. Otherwise, if the provided string is too short, `memcpy()` reads past its end, which is undefined behavior.

I tested with `Serial.println(DateTime("2020-06-29").timestamp());` and it printed `2020-06-29T52:00:253`.

gounselor: You are right, i overlooked that. Do you open a pull request?

edgar-bonet: OK, will do.

gounselor: Merci.

Issue #179, opened 2020-07-25
Use of String class can be unreliable for Arduino


GunZi200:
Hi I just wanted to report an issue I had and how I fixed it. I'm very curious if someone can point out what I was doing wrong because editing the library feels like a dirty fix.

#### IDE: 
 * Visual Studio Code with Platform IO plugin (latest 4.4.0a7)

#### Hardware:
  * Arduino Mega
  * Arduino Ethernet Shield 2
  * DS1307 RTC (I2C interface)
  * Modem for internet connection

In short I am publishing data using MQTT through the Ethernet shield and using DS1307 to timestamp my data with format `DateTime::TIMESTAMP_FULL`.

Most of the time the program would abruptly freeze and at first I thought it must be an overflow somewhere (I want to note that the program/ATmega does not reset which can sometime happen when overflow occurs). I isolated it to the following line:

```c
(rtc.now().timestamp(DateTime::TIMESTAMP_FULL)).toCharArray(time_buf, 20); // <-- program hangs here
```

```c
void mqtt_publish(char* topic, float data) {
    // 1. Convert float to char*
    // 2. Get timestamp and convert it to data type char*
    // 3. Create payload
    // 4. Publish data
    // 5. Check for error is returned

    // Step 1 
    char data_buf[5];
    dtostrf(data, 5, 2, data_buf);

    char time_buf[20]; // buffer for time stamp

    // Step 2
    (rtc.now().timestamp(DateTime::TIMESTAMP_FULL)).toCharArray(time_buf, 20); // <-- program hangs here

    // Step 3
    static char *payload = (char * ) malloc(25 * sizeof(char)); // allocate memory for MQTT message.
    uint8_t message_length = (size_t) payload;                  // length of MQTT message.
    sprintf(payload, "%s ppm %s", data_buf, time_buf);          // create message.

    // Step 4 (Retain = false, QOS  = 1)
    uint8_t error = client.publish(topic, payload, message_length, false, 1);

    free(payload); // free memory

    // Step 5 (not shown since its not relevant)
}
```

All in all this is my currently working function after modifying the library:

```c
void mqtt_publish(char* topic, float data) {
    // 1. Convert float to char*
    // 2. Get timestamp and convert it to data type char*
    // 3. Create payload
    // 4. Publish data
    // 5. Check for error is returned

    // Step 1 
    char data_buf[5];
    dtostrf(data, 5, 2, data_buf); 

    // Step 2
[...]

PR #180, opened 2020-07-25
Remove use of String in timestamp function. Use char array instead.

GunZi200:
Fixes issue #179 

The `String` class is only used in the library within the `timestamp()` function. It can be unreliable on the Arduino (AVR) platform. I think this change could also be about consistency, why not let the end user choose whether he uses String class or not? 

The code should behave just like the unmodified function. I only changed the data type from String to char *. As I understand it char * should have less limitations than String.

Only the example Timestamp needs to be modified. I can do that if this change is OK.

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -689,9 +689,11 @@ bool DateTime::operator==(const DateTime &right) const {
     (`TIMESTAMP_FULL`).
 
     @see The `toString()` method provides more general string formatting.
-
+    
+    @param buffer A char array with 20 bytes. The number 20 is specified to
+    ensure the input buffer is never too small. The method will overwrite
+    the given buffer with the formatted date and/or time specified by opt.
     @param opt Format of the timestamp
-    @return Timestamp string, e.g. "2020-04-16T18:34:56".
 */
 /**************************************************************************/
 void DateTime::timestamp(char buffer[20], timestampOpt opt) {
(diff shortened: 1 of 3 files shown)


GunZi200:
Output from the updated example using Visual Studio Code + PlatformIO
image

edgar-bonet:
This is called a _breaking change_: a change in the library's API that makes it incompatible with previous versions. Breaking changes should be avoided in a public library that other people depend on. The users should be able to upgrade their installation of the library without their old programs breaking.

Note that you can _add_ your new `timestamp()` method as an overload, without removing the existing one. Doing so would preserve API compatibility. If you do so, you should probably implement the current String-returning method by calling the new one, in order to avoid code duplication.

> @param buffer A char array with 20 bytes.

You could specify a format-dependent size:

| format          | size |
|-----------------|-----:|
| TIMESTAMP\_DATE |   11 |
| TIMESTAMP\_TIME |    9 |
| TIMESTAMP\_FULL |   20 |

> ```cpp
> void timestamp(char buf[20], timestampOpt opt = TIMESTAMP_FULL);
> ```

I think the compiler will ignore the specified size and take this as a synonym of

```cpp
void timestamp(char *buf, timestampOpt opt = TIMESTAMP_FULL);
```

drak7:
Issue #179 was resolved, it's not necessary to change this.

PR #184, opened 2020-07-30
Improved `operator==()` and `operator<()` efficiency

Harrison-O:
Refactored `operator==()` and `operator<()` methods to use data members directly instead of using `unixtime()` method. Greatly improved efficiency.

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -660,6 +660,9 @@ TimeSpan DateTime::operator-(const DateTime &right) {
 /*!
     @author Anton Rieutskyi
     @brief  Test if one DateTime is less (earlier) than another.
+    @warning if one or both DateTime objects are invalid, logic errors can
+        occur
+    @see `isValid()` method
     @param right Comparison DateTime object
     @return True if the left DateTime is earlier than the right one,
         false otherwise.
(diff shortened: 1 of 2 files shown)


edgar-bonet:
The warning “if one or both DateTime objects are invalid, **logic errors can occur**” (emphases mine) sounds both vague and scary to me. What could these “logic errors” mean? Could the program just do absolutely anything, like ignoring its control logic or resetting itself? This is reminiscent of the spectre of undefined behavior, complete with its proverbial nasal demons.

Truth is, nothing bad will happen. The operators will just return either `true` or `false`, without breaking the innards of the program. I would replace the sentence with something less scary, like:

> If one or both DateTime objects are invalid, the returned value is meaningless.

But then, it sounds like stating the obvious: how could it be meaningful to ask whether a given date is before or after an invalid date? Maybe we could remove the warning, and instead add a note in the documentation of `isValid()`. Something along the lines:

> `isValid()` is the _only_ method that returns a meaningful value when called on an invalid object.

drak7:
Agreed, "logic errors" is a bit vague. I think adding the note on isValid() is better than a note on every comparison operator. If your object is invalid then it's not going to do anything useful.

See the build details for clang's formatting suggestion.

I tested the changes on an Uno and the speedup is significant if you need to eke out every microsecond.

Harrison-O:
If we removed the warning for the comparison operators it's likely someone not familiar with this library may not realise the importance of using `isValid()` or even the possibility of a DateTime object being invalid. They may just use a comparison operator, complain that the library is buggy and get frustrated. With the warning, they'll realise the importance of validating the date/time components without needing prior knowledge of `isValid()`..

PR #185, opened 2020-07-30
Wrote `fixDateTime()` method

Harrison-O:
Wrote `fixDateTime()` method: attempts to fix invalid DateTime object by decrementing invalid time/date components and incrementing next components to compensate. E.g. If DateTime object represents the 32 July 2019 then running `fixDateTime()` will change it to 1 August 2019.

Also wrote `getDaysInMonth()` and `isLeapYear()` functions. These are used by `fixDateTime()` hence the inclusion.

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -146,11 +146,12 @@ uint8_t getDaysInMonth(uint16_t year, uint8_t month) {
   uint8_t days = 0;
 
   if (month == 12) {
-    days = 31;  // needed since daysInMonth does have December days
+    days = 31; // needed since daysInMonth does have December days
   } else if (month < 12) {
     days += pgm_read_byte(daysInMonth + month - 1);
 
-    if (month == 2 && isLeapYear(year)) days++;
+    if (month == 2 && isLeapYear(year))
+      days++;
   }
 
   return days;
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -120,6 +120,17 @@ const uint8_t daysInMonth[] PROGMEM = {31, 28, 31, 30, 31, 30,
 
 /**************************************************************************/
 /*!
+    @brief checks if the year is a leap year
+    @param year The year to checks
+    @return true if a leap year, false otherwise
+*/
+/**************************************************************************/
+bool isLeapYear(uint16_t year) {
+  return year % 400 == 0 || (year % 4 == 0 && year % 100 != 0);
+}
+
+/**************************************************************************/
+/*!
     @brief  Given a date, return number of days since 2000/01/01,
             valid for 2000--2099
     @param y Year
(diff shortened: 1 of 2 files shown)


RTClib.cpp
```diff
@@ -415,6 +454,65 @@ bool DateTime::isValid() const {
...
+  if (hh >= 24) {
+    temp = hh / 24;
+    d += temp;
+    hh -= temp * 24;
+  }
+
+  // make month valid to prevent getDaysInMonth() returning 0
+  // Otherwise, infinite loop possible
+  if (m > 12) {
+    temp = m % 12;
```
edgar-bonet, line 497: This is probably meant to be `m / 12`
```suggestion
    temp = m / 12;
```

RTClib.cpp
```diff
@@ -415,6 +454,65 @@ bool DateTime::isValid() const {
...
+    m -= temp * 12;
+  }
+
+  temp = getDaysInMonth(yOff, m);
+  while (d > temp) {
+    d -= temp;
+    m++;
+    if (m > 12) {
+      yOff++;
+      m--;
```
edgar-bonet, line 508: This should be `m = 1` or `m -= 12`
```suggestion
      m -= 12;
```

edgar-bonet:
I did a small test that shows the effect of the `mm` field overflowing. `fixDateTime()` performs the following conversions:

```text
2000-01-01T00:255:59 (INVALID) -> 2000-01-01T04:15:59 (valid)
2000-01-01T00:255:60 (INVALID) -> 2000-01-01T00:00:00 (valid)
```

The first one is the correct, expected behavior. The second one is incorrect: it should return 2000-01-01T04:16:00.

RTClib.cpp
```diff
@@ -118,6 +118,45 @@ static void write_i2c_register(uint8_t addr, uint8_t reg, uint8_t val) {
...
 
+/**************************************************************************/
+/*!
+    @brief checks if the year is a leap year
+    @param year The year to checks
+    @return true if a leap year, false otherwise
+*/
+/**************************************************************************/
+bool isLeapYear(uint16_t year) {
+  return year % 400 == 0 || (year % 4 == 0 && year % 100 != 0);
```
edgar-bonet, line 129: Within the range of supported years (2000–2099), `year % 4 == 0` is enough. And this test is already used in a [couple](../../blob/1.11.0/RTClib.cpp#L137) of [places](../../blob/1.11.0/RTClib.cpp#L196) in the current code.

RTClib.cpp
Harrison-O: fixed

RTClib.cpp
Harrison-O: fixed

RTClib.cpp
Harrison-O: Simplified `isLeapYear()` and made it static to prevent misuse.

PR #190, opened 2020-08-03
implement iso8601 TimeSpan formatting

### LINKED ISSUE #181: Support ISO 8601 Duration formatting for RTClib's TimeSpan
- Arduino board:  N/a (but I am testing code regularly on an Arduino Mega 2560)
- Arduino IDE version:  1.8.13

This library provides support for ISO 8601 formatted `DateTime`s through the `toString` and `timestamp` methods, but doesn't have a similar option for `TimeSpan`s. ISO 8601 specifies two [formats for durations](https://en.wikipedia.org/wiki/ISO_8601#Durations). (Durations appear to be basically the same thing as this library's `TimeSpan`s)

I would like to add methods that format `TimeSpan`s according to ISO 8601's `PnYnMnDTnHnMnS` or `P<date>T<time>` formats, both a zero allocation version (not sure what to call it, maybe `format`?) and one that returns a `std::string` called `toString`. I would also add a `TimeSpan` constructor taking one argument of `const char *iso8601duration`. Do you think this feature belongs in this library and would you consider merging a PR containing the changes I laid out?

maxbla:
fixes #181 

I ran this test. It basically tests that serializing and constructing returns the same TimeSpan that was serialized.

```c++
int main() {
    std::default_random_engine generator;
    std::uniform_int_distribution<int32_t> distribution(-2147483648, 2147483647);
    char buf[19];
    for (int i=0; i<10000000; i++) {
        int32_t seconds = distribution(generator); 
        TimeSpan t(seconds);
        t.toCharArray(buf, 19);
        TimeSpan t1(buf);
        assert(t1._seconds == t._seconds);
    }
}
```

I was looking for somewhere to put tests like this, but it seems like this repository doesn't really have tests.


RTClib.cpp
```diff
@@ -746,6 +746,130 @@ TimeSpan::TimeSpan(int16_t days, int8_t hours, int8_t minutes, int8_t seconds)
...
+    construct a TimeSpan of 0 seconds. If parsing fails before the end of the
+    string, this constuctor will interpret its argument as the longest
+    well-formed prefix string.
+
+    For example, the string P5M10~ will parse the same as P5M
+
+    @param iso8601 the formatted string
+*/
+TimeSpan::TimeSpan(char* iso8601): _seconds(0) {
+  int32_t negative = 1;
```
edgar-bonet, line 762: `negative` will quite often be positive. Maybe `sign` would be a better name.

RTClib.cpp
```diff
@@ -746,6 +746,130 @@ TimeSpan::TimeSpan(int16_t days, int8_t hours, int8_t minutes, int8_t seconds)
...
+  char* cursor = iso8601;
+  if (*cursor != '\0' && *cursor == '-') {
```
edgar-bonet, line 764: The first test is redundant:
```suggestion
  if (*cursor == '-') {
```
Same for all the similar tests below.

RTClib.cpp
```diff
@@ -746,6 +746,130 @@ TimeSpan::TimeSpan(int16_t days, int8_t hours, int8_t minutes, int8_t seconds)
...
+    Fails (returning 0) if buf is too small to store the result. buf size 19
+    will never fail.
+    Example: TimeSpan t(32, 23, 54, 11) formats to "P32DT23H54M11S".
+
+    @param buf char array to write output
+    @param len the length of buf
+    @return number of bytes written excluding trailing '\0' or 0 on failure
+*/
+size_t TimeSpan::toCharArray(char* buf, size_t len) const {
+  int32_t tmp = _seconds < 0 ? -_seconds: _seconds;
```
edgar-bonet, line 831: This will overflow if `_seconds` is `INT32_MIN`.

RTClib.cpp
```diff
@@ -746,6 +746,130 @@ TimeSpan::TimeSpan(int16_t days, int8_t hours, int8_t minutes, int8_t seconds)
...
+  int8_t seconds = tmp % SECS_PER_MIN;
+  tmp /= SECS_PER_MIN;
+  int8_t minutes = tmp % MINS_PER_HOUR;
+  tmp /= MINS_PER_HOUR;
+  int8_t hours = tmp % HOURS_PER_DAY;
+  tmp /= HOURS_PER_DAY;
+  int16_t days = tmp;
+  size_t written = 0;
+  if (_seconds < 0) {
+    written += snprintf(buf+written, len-written, "-P");
```
edgar-bonet, line 841: Note that some implementations of `snprintf()` return “the number of characters that would have been written if there were enough space”. This can make `written` larger than `len`.

RTClib.cpp
```diff
@@ -746,6 +746,130 @@ TimeSpan::TimeSpan(int16_t days, int8_t hours, int8_t minutes, int8_t seconds)
...
+  } else {
+    written += snprintf(buf+written, len-written, "P");
+  }
+
+  if (days > 0) {
+    written += snprintf(buf+written, len-written, "%dD", days);
```
edgar-bonet, line 847: `len-written` will have the type `size_t`, which is an unsigned type. If `written` > `len`, the subtraction will wrap around to a very large number, which will lead to `snprintf()` overflowing the buffer.

RTClib.cpp
```diff
@@ -746,6 +746,130 @@ TimeSpan::TimeSpan(int16_t days, int8_t hours, int8_t minutes, int8_t seconds)
...
+  }
+
+  if (hours > 0 || minutes > 0 || seconds > 0) {
```
edgar-bonet, line 850: If `_seconds` is zero, this method prints just “P" which, according to the Wikipedia article, “is not a valid representation for a duration of 0 seconds”.

RTClib.cpp
maxbla: Yeah, and I knew to fix this before marking this draft ready for review. It's fixed now!

RTClib.h
```diff
@@ -231,6 +234,9 @@ class TimeSpan {
 
 protected:
   int32_t _seconds; ///< Actual TimeSpan value is stored as seconds
+  static constexpr int32_t SECS_PER_MIN = 60; ///< Number of seconds in a minute
+  static constexpr int32_t MINS_PER_HOUR = 60; ///< Number of minutes in an hour
+  static constexpr int32_t HOURS_PER_DAY = 24; ///< Number of hours in a day
```
maxbla, line 239: How would you feel about making these either global or public?

edgar-bonet: I have no strong feelings either way: they certainly look like generally useful, but they are kind of trivial, and they would risk collisions in the global namespace. Maybe public would be OK. I guess you should get the opinion of a maintainer: I am just a random contributor.

RTClib.cpp
```diff
@@ -746,6 +746,168 @@ TimeSpan::TimeSpan(int16_t days, int8_t hours, int8_t minutes, int8_t seconds)
...
+    If the string is entirely malformed (doesn't begin with P or -) this will
+    construct a TimeSpan of 0 seconds. If parsing fails before the end of the
+    string, this constuctor will interpret its argument as the longest
+    well-formed prefix string.
+
+    For example, the string P5M10~ will parse the same as P5M
+
+    @param iso8601 the formatted string
+*/
+TimeSpan::TimeSpan(char *iso8601) : _seconds(0) {
```
edgar-bonet, line 761: `iso8601` and `cursor` should be `const char *`

RTClib.cpp
```diff
@@ -746,6 +746,168 @@ TimeSpan::TimeSpan(int16_t days, int8_t hours, int8_t minutes, int8_t seconds)
...
+    cursor++;
+  }
+  if (*cursor == 'P') {
+    cursor++;
+  } else {
+    return;
+  }
+
+  char *rest;
+  long num = strtol(cursor, &rest, 10);
```
edgar-bonet, line 775: If there are no digits, `strtol()` returns 0, and this case is not detected: this constructor parses "PDTHM30S" as 30&nbsp;seconds. Not sure that string is valid, the Wikipedia page is not very clear on this. [The Java doc on `Duration.parse()`](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence-) is more specific: it states (emphasis mine) “The number part of each section must consist of **one or more** ASCII digits.”

RTClib.cpp
```diff
@@ -746,6 +746,168 @@ TimeSpan::TimeSpan(int16_t days, int8_t hours, int8_t minutes, int8_t seconds)
...
+  cursor = rest;
+  if (*cursor == 'D') {
+    _seconds += sign * SECONDS_PER_DAY * num;
```
edgar-bonet, line 778: This can overflow.

RTClib.cpp
```diff
@@ -746,6 +746,168 @@ TimeSpan::TimeSpan(int16_t days, int8_t hours, int8_t minutes, int8_t seconds)
...
+    @param len the length of buf. Must be at least 1.
+    @return number of bytes written excluding trailing '\0' or 0 on failure
+*/
+size_t TimeSpan::toCharArray(char *buf, size_t len) const {
+  size_t written = 0;
+
+  if (_seconds == 0) {
+    written = snprintf(buf, len, "PT0S");
+    if (written >= len) {
+      buf[len - 1] = '\0';
```
edgar-bonet, line 843: The multiple instances of this statement can be factored into a single one at the beginning of the function. Then `snprintf()` should write no more than `len-1-written` bytes in order to not overwrite that `'\0'`.

RTClib.cpp
```diff
@@ -746,6 +746,168 @@ TimeSpan::TimeSpan(int16_t days, int8_t hours, int8_t minutes, int8_t seconds)
...
+      return 0;
+    } else {
+      return written;
+    }
+  }
+
+  int8_t sign = _seconds > 0 ? 1 : -1;
+  int32_t tmp = _seconds;
+  int8_t seconds = sign * (tmp % SECS_PER_MIN);
+  tmp /= SECS_PER_MIN;
```
edgar-bonet, line 853: The sign could be folded into `tmp` at this point, in order to avoid extra multiplications later.
```suggestion
  tmp = sign * (tmp / MINS_PER_HOUR);
```
Maybe add a comment explaining that it is not folded earlier in order to avoid the overflow on `INT32_MIN`.

RTClib.cpp
maxbla: In an earlier revision (before my first commit to this PR) they were. The problem is that `strtol`'s endptr argument is of type `char**`, not `char const**`, so later assigning `rest` to `cursor` is not allowed (Arduino builds with -fpermissive, so it's a warning not an error).

RTClib.cpp
maxbla: Yeah, it can. What should I do about it? Detect overflow, return 0 on failure, and add this failure mode to the docs?

RTClib.cpp
maxbla: It could be folded in. I decided to avoid that optimization to keep it readable.

RTClib.cpp
edgar-bonet: Are you sure it's not allowed? Assignments that remove `const` are not allowed but, if I am not mistaken, adding `const` is OK. According to [cppreference.com](https://en.cppreference.com/w/cpp/language/cv), “reference/pointer to _unqualified_ type can be converted to reference/pointer to `const`”. Also, avr-g++ 7.3.0 (as distributed with Arduino 1.8.13) doesn't emit warnings on this, even with `-Wall -Wextra`.

RTClib.cpp
edgar-bonet: I'm not sure. It seems to me that the overflow cannot be detected without significant extra complexity. :-(

RTClib.cpp
maxbla: I misspoke about the exact nature of the problem. If I make `iso8601`, `cursor` and `rest` `const char*`, the exact error message is
```
RTClib.cpp:803:26: warning: invalid conversion from 'const char**' to 'char**' [-fpermissive]
num = strtol(cursor, &rest, 10);
```

edgar-bonet: Oh, I see. The thing is, `rest` should remain `char *`, without `const`.

RTClib.h
maxbla: Maybe a static (constexpr) global then? That way they won't cause global namespace collisions. This library as-is has the magic number antipattern. Fixing it is out of scope for this PR though.

RTClib.cpp
maxbla: I'm fine with this based on the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle). Also I found [this pdf](https://dotat.at/tmp/ISO_8601-2004_E.pdf) online. Not sure how legit it is.

maxbla:
What's the status of this PR? Should I do something to move it along?

edgar-bonet:
I guess you could solve the two issues raised by GitHub:

> This pull request is still a work in progress  
> This branch has conflicts that must be resolved

The conflict is pretty easy to fix:

```diff
diff --cc RTClib.cpp
index e282548,2ecffb4..3a0ff37
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@@ -694,8 -713,8 +713,8 @@@ bool DateTime::operator==(const DateTim
      @return Timestamp string, e.g. "2020-04-16T18:34:56".
  */
  /**************************************************************************/
 -String DateTime::timestamp(timestampOpt opt) {
 +String DateTime::timestamp(timestampOpt opt) const {
-   char buffer[20];
+   char buffer[25]; // large enough for any DateTime, including invalid ones
  
    // Generate timestamp according to opt
    switch (opt) {
```

I would consider rebasing on top of master, and maybe squashing the last 9 commits together.

As for the “work in progress” status, you should be able to change it to “Ready for review”, when you are ready.

RTClib.cpp
maxbla: But then this library would eagerly overwrite the last byte of the user's array even when they only use the first 5 (of say 19) bytes. This doesn't seem highly problematic, but does violates the least surprise principle. On the other hand, the code bloat is probably significant from all those `if`s. I changed it to use only one if at the end.

RTClib.cpp
maxbla: I folded it in.

maxbla:
This PR is ready for (final) review and I have time to fix any problems identified in said review.

RTClib.cpp
```diff
@@ -765,6 +765,168 @@ TimeSpan::TimeSpan(int16_t days, int8_t hours, int8_t minutes, int8_t seconds)
...
+  this->toCharArray(buf, buflen);
+  return String(buf);
+}
+
+// Equivalent to left - right unless that would underflow, otherwise 0
+//
+// At the time of writing, this is only used in logic in TimeSpan::toCharArray
+// This function's inclusion is very unfortunate, and if it is used anywhere
+// else, should probably use a template
+static size_t sat_sub(size_t left, size_t right) {
```
maxbla, line 850: It feels dirty to add this function in here, but it is the cleanest way I could think to handle the repeated
```C++
if (written >= len) {
    buf[len - 1] = '\0';
    return 0;
}
```

RTClib.cpp
maxbla: How would you feel about using [`__builtin_smull_overflow`](https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html). It is supported in both gcc and clang, but it's non-standard.

edgar-bonet: `__builtin_smull_overflow()` looks nice, but I do not know whether it is supported by all the environments where RTClib can potentially be used.

Maybe a simple fix would be to build the absolute value of the duration as a `uint32_t`. That won't prevent the overflow, but at least it would avoid undefined behavior. Then, at the end of the constructor,

```c++
_seconds = sign * abs_seconds;
```

is guaranteed to wrap correctly to the correct signed value, as long as the result is in the range of an `int32_t`.

edgar-bonet (approved):
LGTM. I hope it gets merged, with or without my suggestions.

RTClib.cpp
```diff
@@ -765,6 +765,168 @@ TimeSpan::TimeSpan(int16_t days, int8_t hours, int8_t minutes, int8_t seconds)
...
+    }
+  }
+
+  if (written >= len) {
+    buf[len - 1] = '\0';
+    return 0;
+  } else {
+    buf[written] = '\0';
+    return written;
+  }
```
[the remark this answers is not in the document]
maxbla: Okay, I removed most lines that adds nul bytes

RTClib.cpp
maxbla: > all the environments where RTClib can potentially be used

Is RTClib used outside the Arduino ecosystem? And if so, can it be used with any compiler other than Clang or GCC? (clang [includes these intrinsics ](https://clang.llvm.org/docs/LanguageExtensions.html#multiprecision-arithmetic-builtins) too)

RTClib.h
maxbla: On second thought, `static constexpr`s won't work in this case because they are in the header. Statics in a header are textually `#include`d, and `static` means limited to a "translation uint", so the constants would still be exposed in the public interface.

Issue #193, opened 2020-09-19
alarm_mode DS3231_A1_Day or DS3231_A2_Day bug


Shuptuu:
Hi,
The "Day" mode for alarm2 or alarm1 of a DS3231 doesn't work (i.e h,m,s and day of the week matching).
And I discovered why:
in the library, days of week are considered from 0 to 6, 0 meaning Sunday.
But according the DS3231 datasheet, they must be from 1 to 7 (page 11 in https://datasheets.maximintegrated.com/en/ds/DS3231.pdf)
Also: "_The day-of-week register increments at midnight. Values that correspond to the day of week are user-defined but
must be sequential (i.e., if 1 equals Sunday, then 2 equals Monday, and so on). Illogical time and date entries result
in undefined operation._"
In other words, setting 0 for the day leads to an undefined operation: there's no automatic incrementation.
But this is exactly which is done: in the "adjust" function, line 1397 in RCTLib.cpp, the day register is always set to 0.
So of course, if you set an alarm in the "day" mode, it will never work cause the day mechanism is not set properly.

The fix (tested and validated):
from line 1391:
```
void RTC_DS3231::adjust(const DateTime &dt) {
  Wire.beginTransmission(DS3231_ADDRESS);
  Wire._I2C_WRITE((byte)DS3231_TIME); // start at location 0
  Wire._I2C_WRITE(bin2bcd(dt.second()));
  Wire._I2C_WRITE(bin2bcd(dt.minute()));
  Wire._I2C_WRITE(bin2bcd(dt.hour()));
  // Wire._I2C_WRITE(bin2bcd(0));
  Wire._I2C_WRITE(bin2bcd(dt.dayOfTheWeek()));
  Wire._I2C_WRITE(bin2bcd(dt.day()));
  Wire._I2C_WRITE(bin2bcd(dt.month()));
  Wire._I2C_WRITE(bin2bcd(dt.year() - 2000));
  Wire.endTransmission();

  uint8_t statreg = read_i2c_register(DS3231_ADDRESS, DS3231_STATUSREG);
  statreg &= ~0x80; // flip OSF bit
  write_i2c_register(DS3231_ADDRESS, DS3231_STATUSREG, statreg);
}
```
and from line 577
```
uint8_t DateTime::dayOfTheWeek() const {
  uint16_t day = date2days(yOff, m, d);
  // return (day + 6) % 7; // Jan 1, 2000 is a Saturday, i.e. returns 6
[...]

PR #194, opened 2020-09-20
fix DS3231_A1_Day and DS3231_A2_Day issue

robsonos:
Hi adafruit,

I modified `DateTime::dayOfTheWeek` and `RTC_DS3231::adjust` to reflect day-of-week register as specified on [DS3231](https://cdn-shop.adafruit.com/product-files/3013/DS3231.pdf) and [DS1307](https://datasheets.maximintegrated.com/en/ds/DS1307.pdf) ([see discussion here](https://github.com/adafruit/RTClib/issues/193)). Unit tested on uno 328p ([test file](https://github.com/adafruit/RTClib/files/5251653/main.zip)).
image

A possible limitation is that [PCF8523](https://www.nxp.com/docs/en/data-sheet/PCF8523.pdf) works with 0 to 6 for day-of-week.

Cheers,

Robson

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -571,12 +571,12 @@ uint8_t DateTime::twelveHour() const {
 /**************************************************************************/
 /*!
     @brief  Return the day of the week.
-    @return Day of week as an integer from 0 (Sunday) to 6 (Saturday).
+    @return Day of week as an integer from 1 (Sunday) to 7 (Saturday).
 */
 /**************************************************************************/
 uint8_t DateTime::dayOfTheWeek() const {
   uint16_t day = date2days(yOff, m, d);
-  return (day + 6) % 7; // Jan 1, 2000 is a Saturday, i.e. returns 6
+  return ((day + 6) % 7) + 1; // Jan 1, 2000 is a Saturday, i.e. returns 7
 }
 
 /**************************************************************************/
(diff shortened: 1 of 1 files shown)


edgar-bonet:
As I stated in a [comment to issue #193][comment], this would be a breaking change, and breaking changes should be avoided on a published library many people rely on.

Would you mind testing the change I proposed in that issue?

[comment]: https://github.com/adafruit/RTClib/issues/193#issuecomment-695784437

robsonos:
Hi @edgar-bonet.

 I like you solution better, thanks. I just tested your fork here and all tests passed.

One more thing, as mentioned before `RTC_DS1307` uses 1-7 as well. That means that `RTC_DS1307::adjust` may need to use `dowToDS3231` as well. I don't have that IC to test here though.

Let me know when you create the PR for your changes.

Cheers.

edgar-bonet:
> I just tested your fork here and all tests passed.

Cool. :-)

> `RTC_DS1307::adjust` may need to use `dowToDS3231` as well

I am not convinced it is worth changing the code for the DS1307, as that RTC doesn't have alarms, and computing the day of the week can take a non-trivial amount of CPU cycles. One use case I can imagine where it would be useful to write the day-of-week register is if RTClib is used to set an RTC which is then read by another library that does rely on that register. But it sounds like a convoluted scenario to me.

> Let me know when you create the PR for your changes.

OK done: PR #195. I added one comment to explain why we are updating this register on this particular model of RTC:

```cpp
// The RTC must know the day of the week for the weekly alarms to work.
```

robsonos:
Great. Cheers.

PR #196, opened 2020-09-23
Force unsigned integer aritmetic for all year 2000 calculations

hlovdal:
Also fix printf formatting characters from %d to %u.

Without this fix, compiling would give the following warnings:

Compiling library "RTClib"
/opt/arduino-1.8.13/hardware/tools/avr/bin/avr-g++ -c -g -Os -Wall -Wextra -std=gnu++11 -fpermissive -fno-exceptions -ffunction-sections -fdata-sections -fno-threadsafe-statics -Wno-error=narrowing -MMD -flto -mmcu=atmega2560 -DF_CPU=16000000L -DARDUINO=10813 -DARDUINO_AVR_MEGA2560 -DARDUINO_ARCH_AVR -I/opt/arduino-1.8.13/hardware/arduino/avr/cores/arduino -I/opt/arduino-1.8.13/hardware/arduino/avr/variants/mega -I.../libraries/Chronos/src -I.../libraries/Time -I.../libraries/RTClib -I.../libraries/ArduinoThread -I.../libraries/SmartLCD -I/opt/arduino-1.8.13/hardware/arduino/avr/libraries/Wire/src .../libraries/RTClib/RTClib.cpp -o /tmp/arduino_build_578150/libraries/RTClib/RTClib.cpp.o
.../libraries/RTClib/RTClib.cpp: In member function 'bool DateTime::operator<(const DateTime&) const':
.../libraries/RTClib/RTClib.cpp:672:23: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   return (yOff + 2000 < right.year() ||
           ~~~~~~~~~~~~^~~~~~~~~~~~~~
.../libraries/RTClib/RTClib.cpp:673:24: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
           (yOff + 2000 == right.year() &&
            ~~~~~~~~~~~~^~~~~~~~~~~~~~~
.../libraries/RTClib/RTClib.cpp: In member function 'bool DateTime::operator==(const DateTime&) const':
.../libraries/RTClib/RTClib.cpp:696:24: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   return (right.year() == yOff + 2000 && right.month() == m &&
           ~~~~~~~~~~~~~^~~~~~~~~~~~~~

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -129,8 +129,8 @@ const uint8_t daysInMonth[] PROGMEM = {31, 28, 31, 30, 31, 30,
 */
 /**************************************************************************/
 static uint16_t date2days(uint16_t y, uint8_t m, uint8_t d) {
-  if (y >= 2000)
-    y -= 2000;
+  if (y >= 2000U)
+    y -= 2000U;
   uint16_t days = d;
   for (uint8_t i = 1; i < m; ++i)
     days += pgm_read_byte(daysInMonth + i - 1);
@@ -225,8 +225,8 @@ DateTime::DateTime(uint32_t t) {
 /**************************************************************************/
 DateTime::DateTime(uint16_t year, uint8_t month, uint8_t day, uint8_t hour,
                    uint8_t min, uint8_t sec) {
-  if (year >= 2000)
-    year -= 2000;
+  if (year >= 2000U)
+    year -= 2000U;
   yOff = year;
   m = month;
   d = day;
@@ -669,8 +669,8 @@ TimeSpan DateTime::operator-(const DateTime &right) {
 */
 /**************************************************************************/
 bool DateTime::operator<(const DateTime &right) const {
-  return (yOff + 2000 < right.year() ||
-          (yOff + 2000 == right.year() &&
+  return (yOff + 2000U < right.year() ||
+          (yOff + 2000U == right.year() &&
            (m < right.month() ||
             (m == right.month() &&
              (d < right.day() ||
@@ -693,7 +693,7 @@ bool DateTime::operator<(const DateTime &right) const {
 */
 /**************************************************************************/
 bool DateTime::operator==(const DateTime &right) const {
-  return (right.year() == yOff + 2000 && right.month() == m &&
+  return (right.year() == yOff + 2000U && right.month() == m &&
           right.day() == d && right.hour() == hh && right.minute() == mm &&
           right.second() == ss);
 }
@@ -724,11 +724,11 @@ String DateTime::timestamp(timestampOpt opt) {
     break;
   case TIMESTAMP_DATE:
     // Only date
-    sprintf(buffer, "%d-%02d-%02d", 2000 + yOff, m, d);
+    sprintf(buffer, "%u-%02d-%02d", 2000U + yOff, m, d);
     break;
   default:
     // Full
-    sprintf(buffer, "%d-%02d-%02dT%02d:%02d:%02d", 2000 + yOff, m, d, hh, mm,
+    sprintf(buffer, "%u-%02d-%02dT%02d:%02d:%02d", 2000U + yOff, m, d, hh, mm,
             ss);
   }
   return String(buffer);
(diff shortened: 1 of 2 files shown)


edgar-bonet:
Note that only lines 672, 673 and 696 needed to be fixed in order to avoid the warning. Everywhere else the `U` suffix is redundant due to the [implicit conversions][conversion] performed by the compiler.

Note also that `%d` is correct in the `printf()` format: the [default conversions applied to variadic arguments][variadic] turn them all into signed ints.

[conversion]: https://en.cppreference.com/w/cpp/language/implicit_conversion
[variadic]: https://en.cppreference.com/w/cpp/language/variadic_arguments#Default_conversions

hlovdal:
Regarding implicit conversions, they are constant source of confusion for beginners (and possibly old timers as well), and improper integer handling is a very common source of security vulnerabilities. Now I cannot see any security related problems by just using `2000` here, but being explicit is a very good habit that helps avoiding them.

Also the code will be more consistent by always using `2000U` in calculations instead of some times `2000` and sometimes `2000U`.

hlovdal:
Regarding conversion, no - there is no type conversion from unsigned int to signed int.

[The C++ standard](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4659.pdf) says under "8.2.2 Function call":

> "When there is no parameter for a given argument, the argument is passed in such a way that the receiving function can obtain the value of the argument by invoking va_arg. ... If the argument has integral or enumeration type that is subject to the integral promotions ... the value of the argument is converted to the promoted type before the call. These promotions are referred to as the default argument promotions."

**If** being the keyword here. Sure smaller types will be widened to int, but `unsigned int` is already the same size as `signed int`,
and has no promotion defined. The standard explicitly states that `unsigned int` and `signed int` have the same rank under "7.15 Integer conversion rank":

> "The rank of any unsigned integer type shall equal the rank of the corresponding signed integer type."


This applies both to C++

```C++
$ cat main2.cpp && g++ -o main2 main2.cpp -Wall -Wextra && ./main2
#include <iostream>

int main(int argc, char *argv[])
{
        char buf[32];

        (void)argc;
        (void)argv;

        unsigned int n = 0xFF00FF00;
        sprintf(buf, "d: %d", n);
        std::cout << buf << std::endl;
        sprintf(buf, "u: %u", n);
        std::cout << buf << std::endl;
}

d: -16711936
u: 4278255360
$
```

as well as C (with slightly different rules, but same result)

```C
$ cat main1.c && gcc -o main1 main1.c -Wall -Wextra && ./main1
#include <stdlib.h>
#include <stdio.h>

int main(int argc, char *argv[])
{
        char buf[32];

        (void)argc;
        (void)argv;

        unsigned int n = 0xFF00FF00;
        sprintf(buf, "d: %d", n);
        puts(buf);
        sprintf(buf, "u: %u", n);
        puts(buf);
}

d: -16711936
u: 4278255360
$
```

edgar-bonet:
> there is no type conversion from unsigned int to signed int [...] smaller types will be widened to int

That's right. However, the variadic arguments passed to `sprintf()` are not `unsigned int`: they are `uint8_t`, a.k.a. `unsigned char`. They are smaller than `int` and subject to integral promotion. The exception being `2000 + yOff`, which is an `int`.

Anyway, the values being positive, it should make no difference. And indeed, avr-libc defines `PRIu8` as `"u"` (probably to forbid the conversion flags only valid for signed numbers), so that should also be valid.

> implicit conversions [...] are constant source of confusion [...] Also the code will be more consistent by always using `2000U`.

Very fair point.

I don't think there is anything wrong with your changes, save for the way they relate to the commit message. My understanding of that message is that the rationale for the commit is:

1. Avoid a compiler warning
2. Fix an error in the choice of the `printf()` conversion specifiers.

And I commented that:

1. From this perspective, many of the changes are not needed
2. The `printf()` conversion specifiers are not wrong.

If the rationale is instead about consistency, and not relying on implicit conversions, then all the changes are perfectly justified. Only the commit message could be better.

hlovdal:
You are correct about "%d" not being wrong for the non-year arguments, I have updated the commit.

hlovdal:
Maintainers, anything missing before this can be completed?

Issue #197, opened 2020-10-09
pcf8523.ino lacks calibration example


Gambalunga:
The example sketch pcf8523.ino lacks a calibration example for the PCF8523 Real_Time Clock.

Note that in the suggested example that I give below the rather complex method of calculating the calibration offset can be reduced to the use of a constant: offset = deviation x constant (where deviation is in seconds/day (gain or loss). The resulting offset must an integer (nearest whole number)

Suggested variation for pcf8523.ino

> // Date and time functions using a PCF8523 RTC connected via I2C and Wire lib
> #include "RTClib.h"
> 
> RTC_PCF8523 rtc;
> 
> char daysOfTheWeek[7][12] = {"Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"};
> 
> void setup () {
>   Serial.begin(57600);
> 
> #ifndef ESP8266
>   while (!Serial); // wait for serial port to connect. Needed for native USB
> #endif
> 
>   if (! rtc.begin()) {
>     Serial.println("Couldn't find RTC");
>     Serial.flush();
>     abort();
>   }
> 
>   if (! rtc.initialized() || rtc.lostPower()) {
>     Serial.println("RTC is NOT initialized, let's set the time!");
>     // When time needs to be set on a new device, or after a power loss, the
>     // following line sets the RTC to the date & time this sketch was compiled
>     rtc.adjust(DateTime(F(__DATE__), F(__TIME__)));
>     // This line sets the RTC with an explicit date & time, for example to set
>     // January 21, 2014 at 3am you would call:
>     // rtc.adjust(DateTime(2014, 1, 21, 3, 0, 0));
>     //
>     // Note: allow 2 seconds after inserting battery or applying external power
>     // without battery before calling adjust(). This gives the PCF8523's
>     // crystal oscillator time to stabilize. If you call adjust() very quickly
>     // after the RTC is powered, lostPower() may still return true.
>   }
> 
>   // When time needs to be re-set on a previously configured device, the
>   // following line sets the RTC to the date & time this sketch was compiled
>   // rtc.adjust(DateTime(F(__DATE__), F(__TIME__)));
>   // This line sets the RTC with an explicit date & time, for example to set
>   // January 21, 2014 at 3am you would call:
>   // rtc.adjust(DateTime(2014, 1, 21, 3, 0, 0));
> 
>   // When the RTC was stopped and stays connected to the battery, it has
[...]

PR #200, opened 2020-10-15
Better documentation for RTC_PCF8523::calibrate()

edgar-bonet:
As discussed in issue #197, the [reference documentation of `RTC_PCF8523::calibrate()`][doc] is too superficial, which makes it impossible to use the method without reading the datasheet. Pull request #199 introduced an example on the usage of this method, which is a very good introduction but does not replace proper reference documentation.

This pull request aims at providing enough information in the reference to make the method usable without having to study the RTC's datasheet.

[doc]: https://adafruit.github.io/RTClib/html/class_r_t_c___p_c_f8523.html#a33e8ae8a0e1ab3e11c4d0dc07c45d1a1

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -1335,15 +1335,34 @@ void RTC_PCF8523::deconfigureAllTimers() {
 
 /**************************************************************************/
 /*!
-    @brief  Use an offset to calibrate the PCF8523.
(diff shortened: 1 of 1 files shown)


Gambalunga:
@edgar-bonet Well done. Thanks for the collaboration 👍

Issue #191, opened 2020-08-04
Fix compiler warnings


maxbla:
if you check the box "show verbose output during compilation" in the Arduino IDE, verifying any library that includes RTClib will print warnings.

```
RTClib.cpp:543:7: warning: 'isPM' may be used uninitialized in this function [-Wmaybe-uninitialized]
       if (isPM) {
       ^
RTClib.cpp:468:28: note: 'isPM' was declared here
   uint8_t hourReformatted, isPM;
                            ^
RTClib.cpp:492:47: warning: 'hourReformatted' may be used uninitialized in this function [-Wmaybe-uninitialized]
         buffer[i + 1] = '0' + hourReformatted % 10;
                                               ^
RTClib.cpp:468:11: note: 'hourReformatted' was declared here
   uint8_t hourReformatted, isPM;
```

It would be better if compiling didn't print warnings, but as far as I can tell, the uninitalized values aren't being read. The warnings are just being overly careful.

Easy fix: change 
```c++
  uint8_t hourReformatted, isPM;
  if (apTag) {     // 12 Hour Mode
    if (hh == 0) { // midnight
      isPM = false;
      hourReformatted = 12;
    } else if (hh == 12) { // noon
      isPM = true;
      hourReformatted = 12;
    } else if (hh < 12) { // morning
      isPM = false;
      hourReformatted = hh;
    } else { // 1 o'clock or after
      isPM = true;
      hourReformatted = hh - 12;
    }
  }
```
to
```c++
  uint8_t hourReformatted = hh;
  bool isPM = hh >= 12;
  if (apTag) {     // 12 Hour Mode
    if (hh == 0) { // midnight
      hourReformatted = 12;
    } else if (hh <= 12) { // morning/noon
      hourReformatted = hh;
    } else { // 1 o'clock or after
      hourReformatted = hh - 12;
    }
  }
```
I can submit a PR making this change, but I'd prefer to wait until my first PR, #190 goes through.

netdudeuk:
I get a variation of these warnings when using PlatformIO for the NODEMCUV2 -


C:\Users\me\.platformio\lib\RTClib\RTClib.cpp: In member function 'char* DateTime::toString(cha
C:\Users\me\.platformio\lib\RTClib\RTClib.cpp:543:7: warning: 'isPM' may be used uninitialized 
       if (isPM) {
       ^
[...]

PR #205, opened 2020-11-08
Fix uninitialized variable warning, issue #191

jrw429:
Simple fix.


drak7:
Check clang-format, same change merged in #211

PR #212, opened 2020-11-23
PCF8563 Alarm Interupt

ldab:
Configure and enabled PCF8563 Alarm Interupt

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -1532,16 +1532,16 @@ void RTC_PCF8563::writeSqwPinMode(Pcf8563SqwPinMode mode) {
 /**************************************************************************/
 void RTC_PCF8563::setAlarm(const DateTime &dt) {
 
-  write_i2c_register(PCF8563_ADDRESS, PCF8563_MINUTE_ALARM, 
+  write_i2c_register(PCF8563_ADDRESS, PCF8563_MINUTE_ALARM,
                      bin2bcd(dt.minute()));
-  write_i2c_register(PCF8563_ADDRESS, PCF8563_HOUR_ALARM,   bin2bcd(dt.hour()));
-  write_i2c_register(PCF8563_ADDRESS, PCF8563_DAY_ALARM,    bin2bcd(dt.day()));
+  write_i2c_register(PCF8563_ADDRESS, PCF8563_HOUR_ALARM, bin2bcd(dt.hour()));
+  write_i2c_register(PCF8563_ADDRESS, PCF8563_DAY_ALARM, bin2bcd(dt.day()));
 
-  uint8_t control_status_2 = 
+  uint8_t control_status_2 =
       read_i2c_register(PCF8563_ADDRESS, PCF8563_CONTROL_2);
 
   // Enable alarm interrupt AIE
-  write_i2c_register(PCF8563_ADDRESS, PCF8563_CONTROL_2, 
+  write_i2c_register(PCF8563_ADDRESS, PCF8563_CONTROL_2,
                      (control_status_2 | 0x02 )); // 00000010b
 
 }
@@ -1561,11 +1561,11 @@ void RTC_PCF8563::setAlarm(uint8_t hour = 0, uint8_t min = 0) {
   // disable day, just in case
   write_i2c_register(PCF8563_ADDRESS, PCF8563_DAY_ALARM, 0x80); // 10000000b
 
-  uint8_t control_status_2 = 
+  uint8_t control_status_2 =
       read_i2c_register(PCF8563_ADDRESS, PCF8563_CONTROL_2);
 
   // Enable alarm interrupt AIE
-  write_i2c_register(PCF8563_ADDRESS, PCF8563_CONTROL_2, 
+  write_i2c_register(PCF8563_ADDRESS, PCF8563_CONTROL_2,
                      (control_status_2 | 0x02 )); // 00000010b
 }
 
@@ -1577,10 +1577,10 @@ void RTC_PCF8563::setAlarm(uint8_t hour = 0, uint8_t min = 0) {
 /**************************************************************************/
 void RTC_PCF8563::disableAlarm(void) {
 
-  uint8_t control_status_2 = 
+  uint8_t control_status_2 =
       read_i2c_register(PCF8563_ADDRESS, PCF8563_CONTROL_2);
 
-  write_i2c_register(PCF8563_ADDRESS, PCF8563_CONTROL_2, 
+  write_i2c_register(PCF8563_ADDRESS, PCF8563_CONTROL_2,
                      (control_status_2 | 0xFD )); // 11111101b
 
 }
@@ -1594,9 +1594,10 @@ void RTC_PCF8563::disableAlarm(void) {
 /**************************************************************************/
 boolean RTC_PCF8563::alarmFired(void) {
 
-  uint8_t control_status_2 = read_i2c_register(PCF8563_ADDRESS, PCF8563_CONTROL_2);
+  uint8_t control_status_2 =
+      read_i2c_register(PCF8563_ADDRESS, PCF8563_CONTROL_2);
 
-  return ( (control_status_2 >> 3) & 1 );
+  return ((control_status_2 >> 3) & 1 );
 }
 
 /**************************************************************************/
@@ -1607,10 +1608,10 @@ boolean RTC_PCF8563::alarmFired(void) {
 /**************************************************************************/
 void RTC_PCF8563::clearAlarm(void) {
 
-  uint8_t control_status_2 = 
+  uint8_t control_status_2 =
       read_i2c_register(PCF8563_ADDRESS, PCF8563_CONTROL_2);
   
-  write_i2c_register(PCF8563_ADDRESS, PCF8563_CONTROL_2, 
+  write_i2c_register(PCF8563_ADDRESS, PCF8563_CONTROL_2,
                      (control_status_2 | 0xF7 )); // 11110111b
 }
 // END RTC_PCF8563 implementation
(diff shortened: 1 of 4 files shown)


edgar-bonet:
This can be a nice addition. I find, however, the overloaded API of `RTC_PCF8563::setAlarm()` a bit clumsy. What about something more in line with the already-implemented alarms? Like:

```c++
/** Alarm modes of the PCF8563 */
enum Pcf8563AlarmMode {
  PCF8563_Alarm_Daily,  /**< daily, when hour and minute match */
  PCF8563_Alarm_Weekly, /**< weekly, when dayOfTheWeek, hour and minute match */
  PCF8563_Alarm_Monthly /**< monthly, when day, hour and minute match */
};

class RTC_PCF8563 {
  // ...
  void setAlarm(const DateTime &dt, Pcf8563AlarmMode alarm_mode);
}
```

Maybe a `clearAlarm()` or `diableAlarm()` method would be a nice to have too.

ldab:
Thanks fort the input `diableAlarm()` is indeed a good candidate.

Your comments about the `RTC_PCF8563::setAlarm( )` does not seem to apply, PCF8563 have a different implementation of the Alarm, where one would set the weekday, day, hour, minute to exact match. Therefore I do not see how `Pcf8563AlarmMode ` would fit here.

edgar-bonet:
> PCF8563 have a different implementation of the Alarm, where one would set the weekday, day, hour, minute to exact match [...]

These are low-level implementation details the user should not have to care about. The “alarm mode” of the PCF8563 is the setting of the `AE_*` bits in the alarm registers: setting any of those registers to `0x80` will prevent it to be considered for deciding whether the current time matches the set alarm.

Here is an untested implementation of `RTC_PCF8563::setAlarm()` that uses the above `Pcf8563AlarmMode` parameter:

```c++
void RTC_PCF8563::setAlarm(const DateTime &dt, Pcf8563AlarmMode alarm_mode) {
  uint8_t minute = bin2bcd(dt.minute());
  uint8_t hour = bin2bcd(dt.hour());
  uint8_t day =
      (alarm_mode == PCF8563_Alarm_Monthly) ? bin2bcd(dt.day()) : 0x80;
  uint8_t weekDay =
      (alarm_mode == PCF8563_Alarm_Weekly) ? bin2bcd(dt.dayOfTheWeek()) : 0x80;

  write_i2c_register(PCF8563_ADDRESS, PCF8563_MINUTE_ALARM, minute);
  write_i2c_register(PCF8563_ADDRESS, PCF8563_HOUR_ALARM, hour);
  write_i2c_register(PCF8563_ADDRESS, PCF8563_DAY_ALARM, day);
  write_i2c_register(PCF8563_ADDRESS, PCF8563_WEEKDAY_ALARM, weekDay);

  // Enable alarm interrupt: AIE bit of Control_status_2.
  uint8_t control_status_2 =
      read_i2c_register(PCF8563_ADDRESS, PCF8563_CONTROL_2);
  write_i2c_register(PCF8563_ADDRESS, PCF8563_CONTROL_2,
                     control_status_2 | 0x02);
}
```

(formatted with clang-format: it makes the repo's CI script happy).

ldab:
After tidying it up and thinking about your suggestion I have few comments as I'm not 100% in favour of the current approach.

1. If an alarm is required on Wednesday 11:00 (aka weekly), using `setAlarm(const DateTime &dt, Pcf8563AlarmMode alarm_mode)` sounds counter intuitive as full DateTime will need to be provided.
1.1. On that case I would suggest `setAlarm(const char *weekday, uint8_t hour = 0, uint8_t min = 0)`, in order to differentiate from `setAlarm(uint8_t day, uint8_t hour = 0, uint8_t min = 0)`
1.2. I guess the same applies for a monthly Alarm, since providing the year and month is not necessary.

Or should we stick with `setAlarm(const DateTime &dt, Pcf8563AlarmMode alarm_mode)` ?

edgar-bonet:
> If an alarm is required on Wednesday 11:00 (aka weekly), using `setAlarm(const DateTime &dt, Pcf8563AlarmMode alarm_mode)` sounds counter intuitive as full DateTime will need to be provided.

It is somewhat inefficient to provide the full `DateTime` (6 bytes) when 3 bytes would be enough to store that information. It is also unintuitive (I realize this now) as the `DateTime` class has no constructor that takes a day-of-week parameter, so you would have to choose an arbitrary Wednesday to stand for the concept of “Wednesday”:

```c++
// 2020-11-25 is a Wednesday.
rtc.setAlarm(DateTime(2020, 11, 25, 11, 0), PCF8563_Alarm_Weekly);
```

> On that case I would suggest `setAlarm(const char *weekday, uint8_t hour = 0, uint8_t min = 0)`, in order to differentiate from `setAlarm(uint8_t day, uint8_t hour = 0, uint8_t min = 0)`

But then there is a string to parse. Textual data is useful for I/O, but it would be an inefficient way to pass data between an application and a library. Maybe an enum? Something like:

```c++
rtc.setAlarm(EVERY_WEDNESDAY, 11, 0):
```

> I guess the same applies for a monthly Alarm, since providing the year and month is not necessary.

Indeed. The “seconds” field is also unused.

> Or should we stick with `setAlarm(const DateTime &dt, Pcf8563AlarmMode alarm_mode)`?

It is less than ideal, but it provides consistency with the way DS3231 alarms are configured. And by using `DateTime`, it avoids having multiple ways of specifying a date.

Maybe combine this with some convenience inline overloads? Like:

```c++
void setAlarm(const DateTime &dt, Pcf8563AlarmMode alarm_mode);
void setAlarm(uint8_t hour, uint8_t min = 0) {
    setAlarm(DateTime(2000, 1, 1, hour, min), PCF8563_Alarm_Daily);
}
void setAlarm(DayOfWeek dow, uint8_t hour, uint8_t min) {
    setAlarm(DateTime(2000, 1, 2+dow, hour, min), PCF8563_Alarm_Weekly);
}
void setAlarm(uint8_t day, uint8_t hour, uint8_t min) {
    setAlarm(DateTime(2000, 1, day, hour, min), PCF8563_Alarm_Monthly);
}
```

Do you think an hourly alarm could be useful too?

edgar-bonet:
I wrote:
> Do you think an hourly alarm could be useful too?

I just stumbled upon some evidence that it may indeed be useful: [Run method Every Hour Passes][stackexchange].

[stackexchange]: https://arduino.stackexchange.com/questions/79550

ldab:
> I just stumbled upon some evidence that it may indeed be useful: [Run method Every Hour Passes](https://arduino.stackexchange.com/questions/79550).

Hourly alarm is not built-in to PCF8563, at least not called "Alarm", on that case the application will need to reset the flag and configure a new hour alarm every interrupt, I would treat it as an application implementation.

ldab:
I thought about it, but I believe the following would conflict as `DayOfWeek` would ultimately enumerate a uint8_t as `uint8_t day`, no?

```c++
void setAlarm(DayOfWeek dow, uint8_t hour, uint8_t min) {
    setAlarm(DateTime(2000, 1, 2+dow, hour, min), PCF8563_Alarm_Weekly);
}
void setAlarm(uint8_t day, uint8_t hour, uint8_t min) {
    setAlarm(DateTime(2000, 1, day, hour, min), PCF8563_Alarm_Monthly);
}
```

edgar-bonet:
An `enum` should not conflict with an `uint8_t`. I did a small test and it appears that, for the purpose of overload resolution, the compiler has no issues telling apart an `enum` from the underlying integer type. Or even telling apart different `enum` types.

ldab:
I wrote it, but will not have the hardware to test for another 3 weeks

RTClib.cpp
```diff
@@ -1519,6 +1519,164 @@ void RTC_PCF8563::writeSqwPinMode(Pcf8563SqwPinMode mode) {
...
+/**************************************************************************/
+/*!
+    @author Leonardo Bispo
+    @brief  Set Weekly Alarm values mode on the PCF8563
+    @param dow  Day of the week to trigger the alarm
+    @param hour Hour to trigger the alarm
+    @param min  Minute to trigger the alarm
+*/
+/**************************************************************************/
+void RTC_PCF8563::setAlarm(DayOfWeek dow, uint8_t hour, uint8_t min) {
```
edgar-bonet, line 1567: gcc says:
```text
warning: unused parameter 'hour'
warning: unused parameter 'min'
```

RTClib.cpp
```diff
@@ -1519,6 +1519,164 @@ void RTC_PCF8563::writeSqwPinMode(Pcf8563SqwPinMode mode) {
...
+    break;
+  case Fridays:
+    t = 1483660800;
+    break;
+  case Saturdays:
+    t = 1483747200; // Saturday, January 7, 2017 0:00:00
+    break;
+  }
+
+  setAlarm(DateTime(t), PCF8563_Alarm_Weekly);
```
edgar-bonet, line 1594: This `DateTime` constructor is computationally expensive. It is much cheaper to:
```c++
setAlarm(DateTime(2017, 1, 1 + dow, hour, min), PCF8563_Alarm_Weekly);
```

RTClib.cpp
```diff
@@ -1559,39 +1560,14 @@ void RTC_PCF8563::setAlarm(const DateTime &dt, Pcf8563AlarmMode alarm_mode) {
 /*!
     @author Leonardo Bispo
     @brief  Set Weekly Alarm values mode on the PCF8563
-    @param dow  Day of the week to trigger the alarm
+    @param dow  Day of the week to trigger the alarm, Sunday is 0
```
edgar-bonet, line 1563: “Sunday is 0” is misleading. If the user provides `0` (or any other integer) as the first argument, the overload will be resolved by the compiler to the method `RTC_PCF8563::setAlarm(uint8_t day, uint8_t hour, uint8_t min)`.

ldab: 6e593953ab9cf0ddb627eaa577a4070afe9f53f3

PR #214, opened 2020-11-24
Correction to comment and debug string.

dodohand:
This pull request fixes a comment and debug string in the DS3231 example code 
to match the code in calculating a timestamp in the future.
Tested and working with Feather M4 Express and DS3231 RTC Feather


edgar-bonet:
Good catch! But it's actually 7 days, **12 hours**, 30 minutes and 6 seconds.

dodohand:
You're so right. Fixed my fix to be even more right.

PR #222, opened 2021-03-03
changes macro to support AVR ATtiny84 with TinyWireM

edslatt:
**Scope of change**
* this is a one-line change that opens the library to support the AVR ATtiny84 as well the AVR ATtiny85
* it updates the RTClib.cpp macros to use TinyWireM in the case of AVR ATtiny85 or AVR ATtiny84
* I tested this library with the AVR ATtiny84 / DS3231 in [EHA](https://gitlab.com/eslatt/escalator-home-automation)

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -63,9 +63,6 @@
 #endif
 
 #if (ARDUINO >= 100)
-#include <Arduino.h> // capital A so it is error prone on case-sensitive filesystems
-// Macro to deal with the difference in I2C write functions from old and new
-// Arduino versions.
 #define _I2C_WRITE write ///< Modern I2C write
 #define _I2C_READ read   ///< Modern I2C read
 #else
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -42,7 +42,7 @@
 */
 /**************************************************************************/
 
-#ifdef __AVR_ATtiny85__
+#if defined(__AVR_ATtiny84__) || defined(__AVR_ATtiny85__)
 #include <TinyWireM.h>
 #define Wire TinyWireM
 #else
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -42,14 +42,14 @@
 */
 /**************************************************************************/
 
-#if defined(__AVR_ATtiny84__) || defined(__AVR_ATtiny85__)
+#include "RTClib.h"
+#if defined(__AVR__) && !defined(TWCR) && defined(USICR)
 #include <TinyWireM.h>
 #define Wire TinyWireM
 #else
 #include <Wire.h>
 #endif
 
-#include "RTClib.h"
 #ifdef __AVR__
 #include <avr/pgmspace.h>
 #elif defined(ESP8266)


edgar-bonet:
Nice idea! I suggest making the test more generic:

```c
#if defined(__AVR__) && !defined(TWCR) && defined(USICR)
```

This will replace Wire.h by TinyWireM.h if the following conditions are all met:

1. We are compiling for an AVR (revealed by the `__AVR__` macro)
2. this MCU does _not_ have a TWI (“Two Wire Interface”, the AVR name for I2C, required by Wire on AVR; `TWCR` is the “TWI Control Register”)
3. it has an USI (Universal Serial Interface, needed by TinyWireM; `USICR` is the “USI Control Register”).

These conditions are met by the ATtiny84 and the ATtiny85, but also many others MCUs.

edslatt:
This sounds great!  Very fast too.  Thanks for your work and expertise @edgar-bonet !

edslatt:
Testing results of `#if defined(__AVR__) && !defined(TWCR) && defined(USICR)` in my `__AVR_ATtiny84__` based build:
* ` __AVR__` is defined as expected
* `TWCR` is not defined as expected
* **but** `USICR` is **not** defined, which was not expected
  * `USICR` is describe on [page 117 of the ATtiny24/44/84 datasheet](https://ww1.microchip.com/downloads/en/DeviceDoc/doc8006.pdf#page=117)

Any ideas on how I should proceed?

ladyada:
TBH, id rather just drop tinywirem support - we have a Wire version for our attiny board support package. this library may not even fit on small chips and the number of folks using tiny chips has gone down as low cost cortex m0 boards have proliferated...

edslatt:
I really hope we can get this one line change in to support the ATtiny84.  I've got a significant amount of time into my project that uses an ATtiny84 with a DS3231.

It looks like USICR is a variable and not a macro, which is why the defined(USICR) check is failing.  I haven't yet found how to test in a macro whether a variable exists.

ladyada:
its ok to keep...just warning that long term, tinywirem is unlikely to be supported!

edgar-bonet:
`USICR` is a “variable-like” macro defined by the avr-libc as something equivalent to

```c
#define USICR (* (volatile uint8_t *) 0x2d)
```

However, the definition comes a little bit too late, from the chain of inclusions: RTClib.cpp → RTClib.h → Arduino.h → avr/io.h → avr/iotn84.h → avr/iotnx4.h. For the test on `USICR` to work, it should come after including RTClib.h. For example (needs to be tested):

```diff
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -42,14 +42,14 @@
 */
 /**************************************************************************/
 
-#ifdef __AVR_ATtiny85__
+#include <RTClib.h>
+
+#if defined(__AVR__) && !defined(TWCR) && defined(USICR)
 #include <TinyWireM.h>
 #define Wire TinyWireM
 #else
 #include <Wire.h>
 #endif
-
-#include "RTClib.h"
 #ifdef __AVR__
 #include <avr/pgmspace.h>
 #elif defined(ESP8266)
```

I guess the subsequent conditional inclusion of either Arduino.h or WProgram.h can be dropped, together with the support for Arduino pre-1.0, as RTClib.h unconditionally includes Arduino.h anyway.

edslatt:
I moved the RTClib.h include up, tested against my __AVR_ATtiny84__ based build and committed the changed code.
I also pulled the superfluous Arduino.h include and tested that against my `ARDUINO >= 100` based build.

I did not remove the WProgram.h include because I'm unprepared to test against `ARDUINO < 100`.

edgar-bonet:
The CI test failed, but it seems the issue is unrelated to the pull request: it's apt failing to install a package.

edslatt:
I also think it's unrelated.  I re-triggered the job with a contrived commit hoping it was intermittent but still no joy.

edslatt:
I triggered a re-run of the CI test and it passed so looks like the issue has been resolved.

edslatt:
@edgar-bonet I'm not sure about the process here.  Is there anything else I can do to help get this pull request executed?  Thank you in advance.

edgar-bonet:
> Is there anything else I can do to help get this pull request executed?

Nothing that I am aware of...

PR #224, opened 2021-03-12
abort() in line 18 resets esp8266

Pius171:
line 18 "abort()" send esp8266 into endless reset loop. removing it solved the problem.
I have only tested this with node mcu esp8266 v1.0

### DIFF
--- a/examples/ds1307/ds1307.ino
+++ b/examples/ds1307/ds1307.ino
@@ -15,7 +15,7 @@ void setup () {
   if (! rtc.begin()) {
     Serial.println("Couldn't find RTC");
     Serial.flush();
-    
+    exit(1);
   }
 
   if (! rtc.isrunning()) {
--- a/examples/ds1307/ds1307.ino
+++ b/examples/ds1307/ds1307.ino
@@ -15,7 +15,7 @@ void setup () {
   if (! rtc.begin()) {
     Serial.println("Couldn't find RTC");
     Serial.flush();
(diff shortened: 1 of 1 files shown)


edgar-bonet:
That statement serves a purpose: to prevent the program from carrying on if the RTC cannot be initialized. If you replace it with `exit(1)`, does that solve the problem you see?

thijstriemstra (approved):
agreed, saw same loop

edgar-bonet:
@thijstriemstra: If you replace `abort()` with `exit(1)`, does that fix the reset loop?

examples/ds1307/ds1307.ino
```diff
@@ -15,7 +15,7 @@ void setup () {
   if (! rtc.begin()) {
     Serial.println("Couldn't find RTC");
     Serial.flush();
-    abort();
+
```
edgar-bonet, line 18:
```suggestion
    exit(1);
```

caternuson:
Thanks. That issue was in *all* the examples :(

Updated all of them in #246 with a slightly different approach of a infinite loop with small delay. The ESP8266 BSP should be nice about its implementation of delay and prevent the WDT reset.

thijstriemstra:
It's good the abort call was removed but the alternative (without any code comment) is a little cryptic (to new users, why is it needed, what does it do) imo:

```
while (1) delay(10);
```

PR #226, opened 2021-04-08
Adding preprocessor variables for custom Wire pins

slyons:
A minimal change to allow specifying custom SDA/SCL pins to pass through to the Wire library.

I'm using the names `RTCLIB_CUSTOM_SDA` and `RTCLIB_CUSTOM_SCL` to really drive home the point that it's outside the normal path.

### DIFF
--- a/RTClib.cpp
+++ b/RTClib.cpp
@@ -805,12 +805,17 @@ static uint8_t bin2bcd(uint8_t val) { return val + 6 * (val / 10); }
 
 /**************************************************************************/
 /*!
-    @brief  Start I2C for the DS1307 and test succesful connection
+    @brief  Start I2C for the DS1307 and test succesful connection.
+      Specify RTCLIB_CUSTOM_SDA and RTCLIB_CUSTOM_SCL for custom Wire pins.
     @return True if Wire can find DS1307 or false otherwise.
 */
 /**************************************************************************/
 boolean RTC_DS1307::begin(void) {
+#if defined(RTCLIB_CUSTOM_SDA) && defined(RTCLIB_CUSTOM_SCL)
+  Wire.begin(RTCLIB_CUSTOM_SDA, RTCLIB_CUSTOM_SCL);
+#else
   Wire.begin();
+#endif
   Wire.beginTransmission(DS1307_ADDRESS);
   if (Wire.endTransmission() == 0)
     return true;


edgar-bonet:
How is the user supposed to defined these macros? By modifying the source code of the library?

slyons:
If the `#define`

> How is the user supposed to defined these macros? By modifying the source code of the library?

I've been using it as such:

```C
#define RTCLIB_CUSTOM_SDA 13
#define RTCLIB_CUSTOM_SCL 14
#include "RTClib.h"
```
I originally added another `.begin` method but then didn't want to add too much duplicate code.

ladyada:
we will not take this, it doesnt work anyways.
please submit a PR where a &Wire object is passed in.

ladyada (changes requested):
we dont allow this kind of adjustment to Wire, you can pass in a &Wire object instead

klkl140:
Ive had the same problem about remapped I2C pins.
How about allowing begin() to have the arguments passed through to Wire.. like this:

in RTClib.h:
class RTC_DS1307 {
public:
  boolean begin(int sda=-1, int scl=-1);
...

in RTClib.cpp
boolean RTC_DS1307::begin(int sda, int scl) {
  Wire.begin(sda, scl);
...

Same for the other chips.
If this is acceptable I will try to make a pull request (will be my first...)

ladyada:
we would not accept it, you can use Wire.setPins() outside of RTClib

klkl140:
But this would work on all I2C connections (?)
In my case I have two I2C, one default and the separate for the RTC.
To pass a &Wire object to RTClib is more complicated on the calling side, Wire already has the possibility to take pins.
How about this one without default values:

RTClib.h:
boolean begin();
boolean begin(int sda, int scl);

RTClib.cpp:
boolean RTC_DS1307::begin() {
    begin(SDA, SCL);  // these are used in Wire as defaults
}
boolean RTC_DS1307::begin(int sda, int scl) {
  Wire.begin(sda, scl);
...

ladyada:
that is not true - **only esp32 does this**. it would break every other board so it would not be merged

klkl140:
ok.... I did not thought this is possible....
so your first advice would be best. Like its done in Adafruit_BMP3XX (?)
I'll give it a try

klkl140:
it is working. But it will change much code and I can only test it for DS1307.
Should we wait for [https://github.com/adafruit/RTClib/pull/215](url)? This will make things easier.

ladyada:
sure!

PR #231, opened 2021-05-15
Added option to initialise lib with custom TwoWire interface on SAMD21 boards

ubidefeo:
Thank you for creating a pull request to contribute to Adafruit's GitHub code!
Before you open the request please review the following guidelines and tips to
help it be more easily integrated:

- This pull request adds the option to pass a custom `TwoWire` instance to an instance of any supported RTC modules.
When using a SAMD21 (or other supporting custom SERCOM) a user might decide to have multiple i2c bus or use an extra one mapped on pins which aren't used. This will allow the user to simply pass in a reference to a `Wire` object and that will be used for any `read`, `write`, `send`, `receive` operations.
It adds an overloaded `bool begin(TwoWire *wireInstance);` method but still allows the simple `bool begin()` to work as expected on the default `Wire` interface.
It also adds a `customWire_DS3231onSAMD21` example and rectifies the paragraph of the `library.properties` for additional Library Manager compliance.

- I have tested the changes on ATMega328 (UNO), ATMega32u4 (Leonardo), SAMD21 (Nano 33 IoT), RP2040 mbed (Nano RP2040 Connect)
It still works as it used to before this PR when the original `begin()` method is used.

- Tests have been carried out, and a new Sketch is added to the examples.

Hope you'll like this PR and find it useful. I know I do.
Thank you
Ubi


ubidefeo:
Tests for custom SAMD21 example fail for unsupported platforms such as AVR and ESP* (no SERCOM resources) and for a couple of SAMD21 based boards:
* Arduino Zero Native
* Trinket M0

If this PR has any chance of making it through I'd like to address the SAMD boards, but for unsupported architectures I'm not sure if adding a series of `#ifdef`, `#if defined()` would be nice in the sketch.
I also don't want to ask for a special test for this example which IMO is very valuable for SAMDX users.
It compiles beautifully on the Metro M4 as well :)

If you're interested let me know and I'll put some time in it
thank you

edgar-bonet:
You may want to add empty files in examples/customWire\_DS3231onSAMD21/ with names such as `.metro_m4.test.only`. One such file for each platform where the build of the example is expected to succeed.

For the list of tested platforms, see the [array `main_platforms` in build\_platform.py][main-p], in the ci-arduino repository.

[main-p]: https://github.com/adafruit/ci-arduino/blob/4e069bd16bed87fe0cc6f8bf12e4fe48291681cc/build_platform.py#L94-L95

ubidefeo:
thank you @edgar-bonet for the hint
I believe it now passes checks

src/RTClib.h
```diff
@@ -390,7 +396,8 @@ enum Pcf8523OffsetMode {
 /**************************************************************************/
 class RTC_PCF8523 {
 public:
-  boolean begin(void);
+  boolean begin();
+  boolean begin(TwoWire *wireInstance);
```
ladyada, line 400: change these to `TwoWire *wireInstance = &Wire` so there is only one begin() function

ladyada (changes requested):

src/RTClib.h
ubidefeo: thank you @ladyada 
I've been around the block a few times and still do silly things 😬

ladyada: thanks :)

src/RTClib.h
```diff
@@ -78,6 +79,7 @@ class TimeSpan;
     inclusive.
 */
 /**************************************************************************/
+static TwoWire *RTCWireBus;
```
drak7, line 82: This only allows a single Wire instance in use. Someone might expect to call begin() with a Wire, then begin() for a different RTC with another hardware bus or soft I2C etc. and be surprised when things don't work.
It'd be more useful to have it as a protected member of each RTC class.

drak7 (changes requested):
See RTClib.h for comments on RTCWireBus.
Check out this page for info on clang-format so that you can check it before submitting: [Formatting with clang-format](https://learn.adafruit.com/the-well-automated-arduino-library/formatting-with-clang-format)

src/RTClib.h
ubidefeo: thanks @drak7 
you're absolutely right.
I'll get to it later today or tomorrow :)

ubidefeo:
hey @drak7 
any chance you can take a look at this? 🙏🏼  😃

drak7 (approved):

NoLANline:
Thanks @ubidefeo, you're PR broke my code; and looks like many others. Thanks again for 2 hours of wasted time. In the future please be considerate of other processors in a multi-processor library.. it's a professional practice really. 

@Anyone receiving compilation error's with RTClib and ATTinyXX, any revision AFTER AND INCLUDING 14.0.0 will NOT work! You must use an earlier version!!!

PR #252, opened 2021-12-10
Add DS3232

jaggil:
The DS3232, an accurate I²C RTC with integrated crystal and SRAM, is similar to the DS3231. However, it has the following modifications:
The DS3232 has 236 bytes of battery-backed SRAM.

The addition of the I²C timeout function. This limits the minimum frequency at which the I²C interface can be operated.

The 32kHz output driver is changed to push-pull. This removes the need for an external pullup resistor, resulting in space savings. This also allows faster edges on the clock, along with power savings in the device.

On battery switchover, the BB32kHz bit can be used to selectively enable/disable the 32kHz output.

The 32kHz output on the DS3232 is designed to drive low when turned off. The DS3231 is switched to a high impedance state when switched off.

On the DS3232, the temperature-conversion rate can now be controlled using 2 CRATE bits. These bits control the sample rate of the device. The sample rate determines how often the temperature sensor makes a conversion and applies compensation to the oscillator. Decreasing the sample rate decreases the overall power consumption by decreasing the frequency at which the temperature sensor operates.


edgar-bonet:
Not reviewing, just a few comments:

1. The CI check that failed is about code formatting rules. As suggested in the README, you may consider using clang-format to automatically ensure compliance with these rules. See [Formatting with clang-format][clang]. Alternatively, you may apply [the patch provided by the failed check][patch].

2. The last commit (732ec6f2777e6549fba08eb060abd21af33ce939) created an extra copy of RTClib.h at the root of the repository. This is most likely a mistake to be reverted.

3. As this PR contains many commits which are fixes to previous commits, I would consider squashing all of them together. That is, unless the maintainers express a preference for keeping them all.

[clang]: https://learn.adafruit.com/the-well-automated-arduino-library/formatting-with-clang-format
[patch]: https://github.com/adafruit/RTClib/runs/4450204067?check_suite_focus=true

PR #259, opened 2022-05-13
Replace `boolean` type by `bool`

edgar-bonet:
The `boolean` type [is being phased out][issue] by the Arduino team in favor of the standard `bool` type. According to [the Arduino documentation of `boolean`][doc]:

> It’s recommended to instead use the standard type `bool`, which is identical.

This pull request replaces every occurrence of `boolean` by `bool`.

[issue]: https://github.com/arduino/Arduino/issues/4673
[doc]: https://www.arduino.cc/reference/en/language/variables/data-types/boolean/


ladyada (approved):
thank u!

PR #292, opened 2023-10-13
Add ability to set halt state on DS1307

AmirHmZz:
This PR adds ability to change halt state on DS1307 using `halt()` method.


edgar-bonet:
I noticed this PR has a failed check. It is clang-format complaining that `RTC_DS1307::halt()` does not follow the standard indentation style of this library, which is two spaces per level. You may fix it by hand or run:

```shell
clang-format -i src/RTC_DS1307.cpp
```

AmirHmZz:
@edgar-bonet Thank you. I've fixed it.

Issue #307, opened 2024-10-13
No support for DST


J-Brinkman:
A major part of Europe is using DST. However the library does not support DST. And implementing DST is not trivial. 
So please add support for DST. If not possible, than add the following code in one of the examples.

/* DST support */
/* CC BY-NC-SA Jeroen Brinkman */
/* Release 1: dd 23-10-2024 */

#include <RTClib.h>   // https://github.com/adafruit/rtclib 
RTC_DS3231 rtc;

#define USEDST  true // Use DST (true or false)

DateTime now;

DateTime getclock() { // Return the (DST adjusted) date and time
  
  DateTime n, b, e;
  
  n = rtc.now();
  b = DateTime(n.year(), 3, 31, 3, 0, 0); b = DateTime(n.year(), 3, 31 - b.dayOfTheWeek(), 1, 0, 0);
  e = DateTime(n.year(), 10, 31, 3, 0, 0); e = DateTime(n.year(), 10, 31 - e.dayOfTheWeek(), 3, 0, 0);

  if (USEDST && (n > b) && (n < e)) n = n + TimeSpan(0,1,0,0); // adjust standard time if within summertime
  return n;
}; 

void setclock(DateTime n) { // if the clock is set during summertime then adjust the clock to standard time

  if (USEDST && (rtc.now() != getclock())) n = n - TimeSpan(0,1,0,0); // if summertime then adjust to the standard time
  rtc.adjust(n); // Set the clock to standard time
}; 

void setup() {
  // initialise the rtc 
  rtc.begin();
  if (rtc.lostPower()) setclock(DateTime(F(__DATE__), F(__TIME__)));
}

void loop() {
  now = getclock(); // read the time from the RTC and adjust for DST
}

edgar-bonet:
> please add support for DST

This has already been discussed in the past, and the maintainers have stated that DST is out of scope for this library. See:

 * #144
 * #36

> CC BY-NC-SA

This library being free software (it's MIT-licensed), I wouldn't expect it to accept non-free contributions.

> ```c++
> b = DateTime(n.year(), 3, 31, 3, 0, 0); b = DateTime(n.year(), 3, 31 - b.dayOfTheWeek(), 1, 0, 0);
> e = DateTime(n.year(), 10, 31, 3, 0, 0); e = DateTime(n.year(), 10, 31 - e.dayOfTheWeek(), 3, 0, 0);
> ```

[...]

PR #308, opened 2024-10-13
Create EU DST example

J-Brinkman:
As discussed in Issue #307


examples/EU DST example
```diff
@@ -0,0 +1,52 @@
+/* EU DST example */
```
edgar-bonet: This deserves a few more explanations, stating for instance that:

1. The RTC is kept in local winter time. This is not obvious at first sight, and is quite non-standard (common practice is to keep the RTC either in UTC or in local time).

2. This code is for CET only. It can work across Europe if one is willing to accept the DST change to be one hour off.

examples/EU DST example
```diff
@@ -0,0 +1,52 @@
...
+/* version 1: dd 23-10-2024 */
+
+#include <RTClib.h>   // https://github.com/adafruit/rtclib 
+RTC_DS3231 rtc; 
+//RTC_DS1307 rtc;
+
+#define USEDST  true // Use DST (true or false)
+
+DateTime now;
```
edgar-bonet: This should be local to `loop()`.

examples/EU DST example
```diff
@@ -0,0 +1,52 @@
...
+
+DateTime dstclock(DateTime n) { // Return the given (DST adjusted) date and time according to DST settings (aespecially for date/time checking)
+  
+  DateTime b, e;
+  
+  b = DateTime(n.year(), 3, 31, 3, 0, 0); b = DateTime(n.year(), 3, 31 - b.dayOfTheWeek(), 1, 0, 0); // Last sunday in march 1:00
+  e = DateTime(n.year(), 10, 31, 3, 0, 0); e = DateTime(n.year(), 10, 31 - e.dayOfTheWeek(), 3, 0, 0); // Last sunday in October 3:00
```
edgar-bonet: You have here two calls to `dayOfTheWeek()`, which is quite an expensive function. You could easily call this only once, and only in March or October: you don't need all this on any other month.

examples/EU DST example
```diff
@@ -0,0 +1,52 @@
...
+DateTime getclock() { // Retrieve the (DST adjusted) date and time
+  
+  DateTime n, b, e;
+  
+  n = rtc.now();
+  b = DateTime(n.year(), 3, 31, 3, 0, 0); b = DateTime(n.year(), 3, 31 - b.dayOfTheWeek(), 1, 0, 0); // Last sunday in march 1:00
+  e = DateTime(n.year(), 10, 31, 3, 0, 0); e = DateTime(n.year(), 10, 31 - e.dayOfTheWeek(), 3, 0, 0); // Last sunday in October 3:00
+
+  if (USEDST && (n > b) && (n < e)) n = n + TimeSpan(0,1,0,0); // adjust standard time if within summertime
+  return n;
```
edgar-bonet: Avoid code duplication:
```suggestion
  return dstclock(rtc.now());
```

examples/EU DST example
```diff
@@ -0,0 +1,52 @@
...
+}; 
+
+void setclock(DateTime n) { // if the clock is set during summertime then adjust the clock to standard time
+
+  if (USEDST && (rtc.now() != getclock())) n = n - TimeSpan(0,1,0,0); // if summertime then adjust to the standard time or
+  if (USEDST && (rtc.now() != dstclock(rtc.now()))) n = n - TimeSpan(0,1,0,0); // if summertime then adjust to the standard time or
```
edgar-bonet: You should not rely on `rtc.now()` here: if the user is calling this it's because the RTC is not set.

examples/EU DST example
J-Brinkman: It is also used in the setup part.

examples/EU DST example
J-Brinkman: It is not duplicated. For both the begin and the end date the last sunday is relevant. So the function should be called twice.

examples/EU DST example
J-Brinkman: Very relevant suggestions. 
Point 1 appears to be the key to make it a very simple procedure. 
Point 2 The hour is now set to 1 and 3. This can be adapted to your local time (for instance at 2 and 4) if applicable. Therefore it is an example. 
I would love to add these point to a comment in the proposed code, but have problems to find out how. (This is the very first pull request I ever do).

examples/EU DST example
J-Brinkman: I don't understand. It is called only twice. For March as well as for October.

examples/EU DST example
J-Brinkman: Valid point!  rtc.adjust(n); should be added to line 36.
(Stil figgering out how to adjust the propsed code)

examples/EU DST example
```diff
@@ -0,0 +1,52 @@
+/* EU DST example */
+/* version 1: dd 23-10-2024 */
```
J-Brinkman: /* version 2: dd 23-10-2024 */

examples/EU DST example
J-Brinkman: [DST support v2.txt](https://github.com/user-attachments/files/17356667/DST.support.v2.txt)
I was not able to find out how to incorporate your valueable review comments into the code. I am very sorry about that! So I included a new version with the changes as a result of your review.

examples/EU DST example
edgar-bonet: > This can be adapted to your local time

I believe a comment explaining how to do so would be of some value.

examples/EU DST example
edgar-bonet: I cannot see the variable `now` used in `setup()`.

examples/EU DST example
edgar-bonet: What I mean is that `getclock()` could be reduced to this:

```c++
DateTime getclock() {
  return dstclock(rtc.now());
}
```

examples/EU DST example
edgar-bonet: For amending a pull request, you can simply push extra commits to your “patch-1” branch.

examples/EU DST example
edgar-bonet: It is always called _twice_.  What I mean is that, most of the year, you don't need to call `dayOfTheWeek()` at all. From the 25th to the 31st of October, you do have to call `dayOfTheWeek()` _once_, but you do not care about the last day of March. Conversely, in March you do not care about the last day of October.

Here is a version of the function (untested) that never calls `dayOfTheWeek()` more than once, and only does so 14 days per year:

```c++
// Return the given DateTime adjusted for DST.
// The provided DateTime is assumed to be in UTC+01:00 (i.e. CET, with no DST correction).
// The result will be in CET, with the DST correction applied if needed.
DateTime dstclock(DateTime n) {
  bool isDst;

  if (!USEDST) {
    isDst = false;
  } else if (n.month() < 3 || (n.month() == 3 && n.day() < 25)) {
    isDST = false;  // winter time before 25 March
  } else if (n.month() == 3) {
    // DST starts on last Sunday of March at 02:00 UTC+01:00
    uint8_t dow = n.dayOfTheWeek();
    isDst = (dow == 0 && n.hour() >= 2) || (dow > 0 && n.day() - dow >= 25);
  } else if (n.month() < 10 || (n.month == 10 && n.day() < 25)) {
    isDst = true;  // summer time before 25 October
  } else if (n.month() == 10) {
    // DST ends on last Sunday of October at 02:00 UTC+01:00
    uint8_t dow = n.dayOfTheWeek();
    isDst = (dow == 0 && n.hour() < 2) || (dow > 0 && n.day() - dow < 25);
  } else {  // n.month() > 10
    isDst = false;  // winter time after October
  }

  if (isDst)
    n = n + TimeSpan(0, 1, 0, 0);

  return n;
}
```

This is more source code but, being an `if`/`else` chain, only one of the inner blocks is executed.

I just noticed that, given that this function expects its input to always be in UTC+01:00 (CET winter time), the DST switch always happens at 02:00.

J-Brinkman: I prefer:
`  b = DateTime(n.year(), 3, 31, 1, 0, 0); // Begin of DST: March 31 1:00 (am)
  if (month(n) = 3) b = DateTime(n.year(), 3, 31 - b.dayOfTheWeek(), 1, 0, 0); // Last sunday in March 1:00 (am)
  e = DateTime(n.year(), 10, 31, 3, 0, 0); // End of DST: October 31 3:00 (am)
  if (month(n) = 10) e = DateTime(n.year(), 10, 31 - e.dayOfTheWeek(), 3, 0, 0); // Last sunday in October 3:00 (am)`
  Much simpler and - since it is an example - more understandable.

examples/EU DST example
J-Brinkman: Thanks, that helped!

examples/EU DST example
J-Brinkman: I think that is not wise to do. 
1. It is a compiler directive and good practice to declare these directives on the top of the code.
2. If you have more complex date/time functionality, it is handy to have this boolean available in the complete program

edgar-bonet: Boolean? I am talking about the variable `now`. I agree that the macro `USEDST` belongs to the top of the file.

examples/EU DST example
edgar-bonet: OK, makes a lot of sense! Note, however, that the time switch should not happen at 01:00 or 03:00, but always at 02:00. This is because, as you have now documented, the RTC is kept in winter time all year long, so this function expects its parameter `n` to be in CET winter time (UTC+01:00). In Europe, DST transitions [always happen at 01:00 UTC][dst].

[dst]: https://en.wikipedia.org/wiki/Summer_time_in_Europe

examples/EU DST example
```diff
@@ -0,0 +1,59 @@
+/* EU DST example */
+/* version 3: dd 14-10-2024 */
```
edgar-bonet: This date format is ambiguous. Prefer the ISO date format: 2024-10-14. Or, better yet, remove the line: the history of this file is in the git repository anyways.

examples/EU DST example
```diff
@@ -0,0 +1,59 @@
...
+#define USEDST  true // Use DST (true or false)
+
+DateTime now; 
+
+DateTime dstclock(DateTime n) { // Return the given (DST adjusted) date and time according to DST settings (for extensive date and time calculations)
+  
+  DateTime b, e;
+  
+  b = DateTime(n.year(), 3, 31, 1, 0, 0); // Begin of DST: set on March 31 1:000 (am)
+  if (month(n) = 3) b = DateTime(n.year(), 3, 31 - b.dayOfTheWeek(), 1, 0, 0); // Begin of DST: adjusted to last sunday in March 1:00 (am) when actual month is March
```
edgar-bonet: This doesn't compile. You mean:

```suggestion
  if (n.month() == 3) b = DateTime(n.year(), 3, 31 - b.dayOfTheWeek(), 1, 0, 0); // Begin of DST: adjusted to last sunday in March 1:00 (am) when actual month is March
```

examples/EU DST example
```diff
@@ -0,0 +1,59 @@
...
+  e = DateTime(n.year(), 10, 31, 3, 0, 0); // End of DST: set on October 31 3:00 (am)
+  if (month(n) = 10) e = DateTime(n.year(), 10, 31 - e.dayOfTheWeek(), 3, 0, 0); // End of DST: adjusted to last sunday in October 3:00 (am) when actual month is October
```
edgar-bonet: Same here:

```suggestion
  if (n.month() == 10) e = DateTime(n.year(), 10, 31 - e.dayOfTheWeek(), 3, 0, 0); // End of DST: adjusted to last sunday in October 3:00 (am) when actual month is October
```

examples/EU DST example
```diff
@@ -0,0 +1,59 @@
...
+
+  if (USEDST && (n > b) && (n < e)) n = n + TimeSpan(0,1,0,0); // adjust to standard time if within summertime
+  return n;
+}; 
+
+void setclock(DateTime n) { // if the clock is set during summertime then adjust the clock to standard time
+
+  if (USEDST && (n != dstclock(n)) n = n - TimeSpan(0,1,0,0); // if summertime then adjust to the standard time
```
edgar-bonet: There are mismatched parentheses here:

```suggestion
  if (USEDST && n != dstclock(n)) n = n - TimeSpan(0,1,0,0); // if summertime then adjust to the standard time 
```

examples/EU DST example
```diff
@@ -0,0 +1,59 @@
...
+
+  if (USEDST && (n != dstclock(n)) n = n - TimeSpan(0,1,0,0); // if summertime then adjust to the standard time 
+  rtc.adjust(n); // Set the clock to standard time
+}; 
+
+void setup() {
+  // initialise the rtc 
+  rtc.begin();
+//  if (rtc.lostPower())  setclock(DateTime(F(__DATE__), F(__TIME__))); // Set date and time for use with the DS3231 RTC
+  if (!rtc.isrunning()) setclock(DateTime(F(__DATE__), F(__TIME__))); // Set date and time for use with the DS1307 RTC
```
edgar-bonet: This doesn't compile: the class `RTC_DS3231` doesn't have a method named `isrunning()`. You probably want to comment-out this line and uncomment the previous one. Alternatively, comment-out the current declaration of `rtc` and uncomment `RTC_DS1307 rtc;`.

examples/EU DST example
```diff
@@ -0,0 +1,59 @@
...
+}
+
+void loop() {
+  now = getclock(); // read the time from the RTC and adjust for DST or
+//  now = dstclock(rtc.now()); // read the time from the RTC and adjust for DST
+}
```
edgar-bonet: It would be nice if this example did something visible, just like all the others do. The simplest would be to initialize the serial port in `setup()`, then:

```c++
void loop() {
  DateTime now = getclock(); // read the time from the RTC and adjust for DST
  Serial.println(now.timestamp());
  delay(1000);
}
```

edgar-bonet (commented):
I just noticed the weird name of the file. It should be renamed in accordance with the [Arduino sketch specification][spec]: avoid the space character, add the extension `.ino`, and move it to a directory with the same name (but without the extension). Also, the word `example` is redundant for a sketch belonging to a collection of examples.

I suggest: `EU DST example` → `DST_Europe/DST_Europe.ino`

This is important for the CI pipeline to build-test your example. You may notice that [the build job simply ignored it][ci]. Had the file been properly named, CI would have warned that it doesn't compile.

[spec]: https://arduino.github.io/arduino-cli/0.20/sketch-specification/
[ci]: https://github.com/adafruit/RTClib/actions/runs/11322096804/job/31482205280?pr=308

examples/EU DST example
J-Brinkman: This time it compiles :-) and a nice demonstration is added.

examples/DST_Europe/DST_Europe.ino
```diff
@@ -0,0 +1,96 @@
...
+int lastsec;
+
+DateTime dstclock(DateTime n) { // Return the given (DST adjusted) date and time according to DST settings (for extensive date and time calculations)
+  
+  DateTime b, e;
+  
+  b = DateTime(n.year(), 3, 31, 1, 0, 0); // Begin of DST: set on March 31 1:000 (am)
+  if (n.month() == 3) b = DateTime(n.year(), 3, 31 - b.dayOfTheWeek(), 1, 0, 0); // Begin of DST: adjusted to last sunday in March 1:00 (am) when actual month is March
+  e = DateTime(n.year(), 10, 31, 3, 0, 0); // End of DST: set on October 31 3:00 (am)
+  if (n.month() == 10) e = DateTime(n.year(), 10, 31 - e.dayOfTheWeek(), 3, 0, 0); // End of DST: adjusted to last sunday in October 3:00 (am) when actual month is October
```
edgar-bonet: As I already explained a couple of times, the DST transitions should happen at 02:00 RTC time, not at 01:00 or 03:00.

examples/DST_Europe/DST_Europe.ino
```diff
@@ -0,0 +1,96 @@
...
+
+  if (USEDST && (n > b) && (n < e)) n = n + TimeSpan(0,1,0,0); // adjust to standard time if within summertime
```
edgar-bonet: The test `n > b` should be `n >= b`, otherwise the transition happens one second too late.

examples/DST_Europe/DST_Europe.ino
```diff
@@ -0,0 +1,96 @@
...
+  
+  n = rtc.now();
+  b = DateTime(n.year(), 3, 31, 1, 0, 0); // Begin of DST: set on March 31 1:000 (am)
+  if (n.month() == 3) b = DateTime(n.year(), 3, 31 - b.dayOfTheWeek(), 1, 0, 0); // Begin of DST: adjusted to last sunday in March 1:00 (am) when actual month is March
+  e = DateTime(n.year(), 10, 31, 3, 0, 0); // End of DST: set on October 31 3:00 (am)
+  if (n.month() == 10) e = DateTime(n.year(), 10, 31 - e.dayOfTheWeek(), 3, 0, 0); // End of DST: adjusted to last sunday in October 3:00 (am) when actual month is October
+
+  if (USEDST && (n > b) && (n < e)) n = n + TimeSpan(0,1,0,0); // adjust to standard time if within summertime
+  return n;
+};
```
edgar-bonet: Replace this function with

```c++
DateTime getclock() { // Retrieve the (DST adjusted) date and time
  return dstclock(rtc.now());
};
```

examples/DST_Europe/DST_Europe.ino
```diff
@@ -0,0 +1,96 @@
...
+
+void setup() {
+  // initialise the rtc 
+  rtc.begin();
+  if (rtc.lostPower()) setclock(DateTime(F(__DATE__), F(__TIME__))); // Set date and time for use with the DS3231 RTC
+  //  if (!rtc.isrunning()) setclock(DateTime(F(__DATE__), F(__TIME__))); // Set date and time for use with the DS1307 RTC
+
+  // initialise the serial port 
+  Serial.begin(BAUD);
+  Wire.begin();
```
edgar-bonet: No need to initialize `Wire` here: this is done by RTClib. Besides, there would be no point in initializing `Wire` _after_ `rtc.begin()`.

examples/DST_Europe/DST_Europe.ino
```diff
@@ -0,0 +1,96 @@
...
+  // Start the DST demonstration
+  setclock(DateTime(F(__DATE__), F(__TIME__))); // Set the clock to compile date and time
+
+  // show standard (winter) time
+  now = rtc.now(); // retreive the winter time
+  Serial.print("     Actual standard time: ");
+  Serial.print(now.timestamp(DateTime::TIMESTAMP_DATE)); Serial.print(" ");  // print the date
+  Serial.println(now.timestamp(DateTime::TIMESTAMP_TIME));                   // print the time 
+
+  // show DST corrected time (only when in DST period, otherwise this equals the standard time)
```
edgar-bonet: Contrary to what the comment says, the lines below are not conditioned on the clock being in DST period.

edgar-bonet (commented):
Thanks for fixing the file name. However, you created a new copy of the file, instead of moving (`git mv`) the old one. Now the old one has to be removed (with `git rm`).

I tested this new version with `RTC_Millis`, adding some printout for the October transition. The serial output is a nice demonstration of the DST functionality! Here is what I got:

```
2024-03-31 00:59:57
2024-03-31 00:59:58
2024-03-31 00:59:59
2024-03-31 01:00:00
2024-03-31 02:00:01
2024-03-31 02:00:02
...
2024-10-27 03:59:57
2024-10-27 03:59:58
2024-10-27 03:59:59
2024-10-27 03:00:00
2024-10-27 03:00:01
2024-10-27 03:00:02
```

This is wrong for a few reasons:

1. The March transition should happen at 02:00 winter time = 03:00 summer time. At lest for CET. This transition at 01:00 winter time would be correct for WET, but this is only a small part of Europe (PT, IE and UK).
2. Even for WET, the transition is one second too late. The clock should jump from 00:59:59 to 02:00:00.
3. The October transition should happen at 03:00 summer time = 02:00 winter time. At least for CET. This transition at 04:00 summer time would be correct for EET (FI, GR, RO, BG...).

Here is what the correct transitions would be for CET:

```
2024-03-31 01:59:57
2024-03-31 01:59:58
2024-03-31 01:59:59
2024-03-31 03:00:00
2024-03-31 03:00:01
2024-03-31 03:00:02
...
2024-10-27 02:59:57
2024-10-27 02:59:58
2024-10-27 02:59:59
2024-10-27 02:00:00
2024-10-27 02:00:01
2024-10-27 02:00:02
```

examples/DST_Europe/DST_Europe.ino
J-Brinkman: ----------------- New start ------------------
     Actual standard time: 2024-10-14 16:21:11
Actual DST corrected time: 2024-10-14 17:21:11 <- corrected because in DST

----------------- New start ------------------
     Actual standard time: 2024-11-14 16:21:11
Actual DST corrected time: 2024-11-14 16:21:11 <- not corrected because not in DST

edgar-bonet (approved):
Looks good to me!

examples/DST_Europe/DST_Europe.ino
J-Brinkman: I realized that this function could be replaced with:

#define getclock() dstclock(rtc.now()) 

Do you want me to make that change?

edgar-bonet: I have a slight preference for a function over a #define, as a function is more type-safe and more C++-idiomatic. But I don't care that much, and I am fine with whatever option you prefer.

J-Brinkman:
> As discussed in Issue #307
Do I do need to do something to get this addition published?

edgar-bonet:
@dhalbert: Hi! Do you think you could find some time to review this pull request? Maybe also the other PRs of this repo? Is there an officially-designated maintainer?

J-Brinkman:
Any news regarding the publishing (date) of the new version of the library?

Or was my change request a total waste of time?

 

Jeroen Brinkman

PR #318, opened 2026-01-06
Default constructors added for DateTime and TimeSpan classes

solamyl:
Hi,
please add the default constructors for the `DateTime` and `TimeSpan` classes (this proposed pull request, it is a simple change).

- I think it would be very helpful for DateTime class because DateTime(uint32_t) does not have trivial implementation. It is always good idea to keep the default contructors as simple as possible, because they may be called very often in some cases.
- I don't think it is really necessary for TimeSpan, but I have added it too.
- Another point is, that the usage of "method's default parameter" consumes extra (program) memory, because every call to the method must "compile-in" the default paramer value into the code, so if the parametr is `long` and you have many calls, then hudreds of bytes will be wasted.
- From the functional point of view, they do the same, but they make the final code faster and shorter.

thank you in advance :-)
-- stepan


solamyl:
hm, it seems that clang-format.py failed. I have tried to mimic your coding style as much as I could, but probably was wrong with 1 linefeed...

PR #319, opened 2026-01-21
Update README.md to add ESP32-S3

SnowFlash383935:
I ran your lib on ESP32-S3 and it works. I decided to write this to README.md


edgar-bonet (commented):
The line on the ESP8266 has the note:

> SDA/SCL default to pins 4 & 5 but any two pins can be assigned as SDA/SCL using Wire.begin(SDA,SCL)

Does the `Wire.begin()` trick also work on the ESP32? If so, you should use the same wording as for the ESP8266.

Also, it would be better to add the ESP32 right after the ESP8266.

PR #320, opened 2026-03-19
Added getAlarmedState() function

CoolEou:
Adds a function that checks the interrupt flag of either of the two alarms on the DS3231. The name of the function could maybe be better, but other than that i haven't touched anything other than adding this function, so everything else should still work as normal, my change only adds a little bit of functionality to the library.

I needed this function to control an indicator LED that indicates to the user whether the alarm is active or not (indicating whether the alarm will go off at the alarm time or not).

### DIFF
--- a/src/RTC_DS3231.cpp
+++ b/src/RTC_DS3231.cpp
@@ -351,6 +351,19 @@ bool RTC_DS3231::alarmFired(uint8_t alarm_num) {
 
 /**************************************************************************/
 /*!
+    @brief  Get the state of the alarm interrupt flag
+        @param 	alarm_num Alarm number
+    @return False if interrupt flag is not set, otherwise true
+*/
+/**************************************************************************/
+bool RTC_DS3231::getAlarmedState(uint8_t alarm_num){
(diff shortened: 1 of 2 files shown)


edgar-bonet:
Note that the [CI failure][] is due to clang-format complaining that there should be a space before the opening brace:

```diff
--- ./src/RTC_DS3231.cpp	(original)
+++ ./src/RTC_DS3231.cpp	(reformatted)
@@ -356,7 +356,7 @@
     @return False if interrupt flag is not set, otherwise true
 */
 /**************************************************************************/
-bool RTC_DS3231::getAlarmedState(uint8_t alarm_num){
+bool RTC_DS3231::getAlarmedState(uint8_t alarm_num) {
   uint8_t alarmedState = read_register(DS3231_CONTROL);
   alarmedState &= (1 << (alarm_num - 1));
   return alarmedState;
```

[CI failure]: https://github.com/adafruit/RTClib/actions/runs/23291336173/job/67727449502?pr=320

justinnewitter:
This is functionally equivalent to my PR that has been pending since 2022 🙈 
https://github.com/adafruit/RTClib/pull/269

